PR #66, opened 2019-08-30
Fix some compiler warnings in the example cluster

andreasstieger:
Fixes some compiler warnings on buffer sizes


freeekanayaka:
@andreasstieger thanks!

PR #112, opened 2020-01-15
Clean up test library

freeekanayaka:
- Rename a few test API to follow current conventions.
- Tune autoconf macros to modify `AM_CFLAGS` not `CFlAGS` (which is reserved for the user)
- Add a test matrix parameter to fuzzy tests to exercise elections both with and without pre-vote


freeekanayaka:
@stgraber @brauner I resumed this branch and it's now ready to land. The most relevant change is adding test matrix parameter to fuzzy tests to exercise elections both with and without pre-vote (a bug was discovered in pre-vote lately).

PR #119, opened 2020-02-23
Cross-platform support

AshfordN:


freeekanayaka:
I assume this is still a work in progress?

AshfordN:
yeah, I'm actually stuck at trying to get this linked with libuv

paulstuart:
@AshfordN The "My" prefix for disambiguation might be better as "Raft". Just a thought.

Would also like to get some darwin love in on these changes but I guess that can happen with a followup after this lands.

freeekanayaka:
What's the status of this branch? Is it ready for review and possibly landing?

freeekanayaka:
> What's the status of this branch? Is it ready for review and possibly landing?

@AshfordN ? ^^^

AshfordN:
Hey,
I apologise for my inactivity on this branch but I've been a bit busy lately. At the moment, I can't seem to get this linked with libuv despite my best efforts. Although I must admit that I'm actually cross-compiling with MinGW on Linux; which might not be the ideal environment for the ground work that I'm doing.
Aside from that, there were some low-level memory handling [here](https://github.com/canonical/raft/blob/master/src/heap.c#L34) that could not be ported 1-to-1 (see [this](https://developercommunity.visualstudio.com/content/problem/468021/c17-stdaligned-alloc%E7%BC%BA%E5%A4%B1.html) article).
I'd appreciate it if someone could reviewed what I have so far and make the necessary recommendations.

freeekanayaka:
> Hey,
> I apologise for my inactivity on this branch but I've been a bit busy lately.

No worries!

> Aside from that, there were some low-level memory handling [here](https://github.com/canonical/raft/blob/master/src/heap.c#L34) that could not be ported 1-to-1 (see [this](https://developercommunity.visualstudio.com/content/problem/468021/c17-stdaligned-alloc%E7%BC%BA%E5%A4%B1.html) article).

The article itself has this comment at the bottom:

> The Universal C Runtime that is provided with Visual Studio via the Windows 10 SDK does not include support for aligned_alloc. Instead, _aligned_malloc and _aligned_free must be used instead. 

So can't use use `_aligned_malloc` as documented [here](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/aligned-malloc?view=vs-2017)?

> I'd appreciate it if someone could reviewed what I have so far and make the necessary recommendations.

Not sure I'll be able to provide much feedback, since I'm not familiar with Windows. From my point of view, as long as the code works on Windows (something which I trust you to test) **and** does not break Linux (something which will be tested automatically by Travis), I've no objection merging your work.

AshfordN:
>So can't you use _aligned_malloc as documented [here](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/aligned-malloc?view=vs-2017)?

It's not as straight forward. I made a short comment in my version of `heap.c` about this issue. Currently all memory is freed with a call to `free()`. However, memory allocated with `_aligned_malloc()` needs to be freed with `_aligned_free()`, so there needs to be extra bookkeeping in order to keep things safe, and I'm not quite sure how to attach that extra bookkeeping at such a low level.
I left it half-done for now since I didn't want to make major changes to the code structure.
I'll attempt to bring this to review stage sometime this week.

freeekanayaka:
> > So can't you use _aligned_malloc as documented [here](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/aligned-malloc?view=vs-2017)?
> 
> It's not as straight forward. I made a short comment in my version of `heap.c` about this issue. Currently all memory is freed with a call to `free()`. However, memory allocated with `_aligned_malloc()` needs to be freed with `_aligned_free()`, so there needs to be extra bookkeeping in order to keep things safe, and I'm not quite sure how to attach that extra bookkeeping at such a low level.
> I left it half-done for now since I didn't want to make major changes to the code structure.
> I'll attempt to bring this to review stage sometime this week.

Ah I see. There are very few cases were aligned_malloc is needed (basically just for direct I/O), it should not be too bad to special-case them and use a dedicated free().

src/uv_os.c
```diff
@@ -37,6 +39,13 @@ int UvOsClose(uv_file fd)
     return uv_fs_close(NULL, &req, fd, NULL);
 }
 
+#ifdef _WIN32
+int UvOsFallocate(uv_file fd, off_t offset, off_t len)
+{
+    //TODO: implement functionality
+    return 0;
+}
+#else
```
Kerollmops, line 48: It seems like `posix_fallocate` is also missing on OSX, I found out that mozilla is faking it by using `ftruncate` under the hood.

https://stackoverflow.com/a/11497568/1941280

AshfordN:
Hey,
My main challenge with this is actually completing the build process. For some reason, MinGW isn't linking libuv and I'm not sure why. Nevertheless, I'm still able to complete the build process when targeting Linux, so I don't think I broke the build script. Work has been preventing me from pinning down the issue so I'd appreciate it if someone more familiar with windows could point out any shortcomings.
As you said, everything seem to be in place to complete the work on this PR, but since I'm unable to test it, I can't say for sure.

@Kerollmops am I to understand that your branch is working on OSX?

AshfordN:
@freeekanayaka also, does my fallback code [here](https://github.com/AshfordN/raft/blob/master/src/uv_writer.c#L533) look sensible?

freeekanayaka:
> @freeekanayaka also, does my fallback code [here](https://github.com/AshfordN/raft/blob/master/src/uv_writer.c#L533) look sensible?

That might be enough in practice, although it does not completely match the current semantics: for example `UvWriterClose` currently waits for in-flight write requests to complete, before firing the given `UvWriterCloseCb`. Also, you'll need to put in place some more code to fire the `UvWriterReqCb` and `UvWriterCloseCb`.

Considering that the current implementation already has fallback support for performing the writes using libuv's threadpool (which is exactly how the `uv_fs_write` API you are using is implemented internally in libuv), then I'd suggest to disable the non-portable code (`eventfd`/`io_submit`) using `#if`/`#else`. You'll need to put in place more `#if`/`#else` dance, but you'll be able to leverage the existing fallback implementation without essentially re-inventing it.

Kerollmops:
> As far as I can tell, with the introduction of raft_heap->aligned_free() and the internal posix_fallocate emulation code there road should now be paved for completing this work.

It seems to do the work but I didn't tried it, I didn't tried to create a wrapper of this API around the Rust Allocator API, sorry. I will neither be able to test that in a near futur, but in my understading this `aligned_free` function was the only missing one.

> @Kerollmops am I to understand that your branch is working on OSX?

I wouldn't say that it works but at least it compiles and kind of run until it locks or something. No more messages are passed between peer. It is really hard to debug this kind of stuff and don't have much time to allocate to it :/

AshfordN:
> I wouldn't say that it works but at least it compiles and kind of run until it locks or something. No more messages are passed between peer.

This might due to the questionable implementation of the fallback code discussed above. The fix might not be too difficult based on what @freeekanayaka said, but I'd have to figure out the build process so that I can at least test the code. Unfortunately I can't give a definite timeline for this PR, but I will try to complete the work.

jonahharris:
Has there been any non-public work on this lately? While I don't have a lot of time, it would be cool to get this running on OSX. I have it compiled and kind-of working... but I definitely don't trust it.

freeekanayaka:
> Has there been any non-public work on this lately?

Not that I'm aware of.

gabriel-samfira (commented):
I managed to cross compile this branch, with a few changes using mingw. It seems to work. The changes I made can be viewed here:

https://github.com/gabriel-samfira/raft/tree/hack

This branch rebases your changes and adds a few changes that were needed to get dqlite to build against it. The dqlite branch (if you're curious), is here:

https://github.com/gabriel-samfira/dqlite/pull/new/hack-it-to-bits

I have to mention that my C skills are extremely lacking. I essentially learned enough C to hack this with an axe in an attempt to get it to build with mingw. The goal was not to make a merge-able PR. So as you go through my branch(es), prepare to cringe...a lot.

ericcurtin:
Interesting PR, tempting to crack open my Windows development environment which has been lying dormant for over a year now. clang is also worth a consideration for windows, it compiles native Windows binaries now (though I think you still need to use a different linker, unless that's changed since I last looked at it). Although mingw is a nice alternative also, clang is more native. If you compile directly on Windows also, it might make it easier to run the test suite on Windows.

Issue #130, opened 2020-04-17
Progress of a Rust wrapper


Kerollmops:
I am currently working on making a good and safe abstraction/wrapper of libraft in Rust that I will probably release under the name `canonical-raft`.

The story of replicated system in Rust is worse than the one in Go where the most popular library to replicate states over the network is the [hashicorp/raft](https://github.com/hashicorp/raft) library. The power of it is mainly comes from the ease of use of it: you only have to implement [one single and simple FSM interface](https://godoc.org/github.com/hashicorp/raft#FSM) and you are done.

I would love to be able to make that easy in Rust, it would highly empower the language!
The only viable Raft library out there is [the one from pingcap](https://github.com/tikv/raft-rs) and I must say that it is not easy to use **at all**!!!

In a second time I would like to create a pure Rust implementation of a `raft_io` backend that I will be able to include with the library. It would be easier to integrate in the async ecosystem of Rust than the `uv` based one and it would be more cross-platform too.
Do you think it could be possible?

The last thing that I have seen is that the hashicorp/raft library comes with [an official `MDBStore` backend](https://github.com/hashicorp/raft-mdb), as I am the author of heed: a safe typed LMDB wrapper in Rust, I will probably make an equivalent store on top of LMDB using heed. 
Does that make sense to you?

I would like to thank you about the great work you have done here!
I was searching for a Raft or Paxos simple library to make our [search engine MeiliSearch](https://github.com/meilisearch/MeiliSearch) replicated for too long now!

freeekanayaka:
> I am currently working on making a good and safe abstraction/wrapper of libraft in Rust that I will probably release under the name `canonical-raft`.
>
> The story of replicated system in Rust is worse than the one in Go where the most popular library to replicate states over the network is the [hashicorp/raft](https://github.com/hashicorp/raft) library. The power of it is mainly comes from the ease of use of it: you only have to implement [one single and simple FSM interface](https://godoc.org/github.com/hashicorp/raft#FSM) and you are done.
> 
> I would love to be able to make that easy in Rust, it would highly empower the language!

Yeah, there's no particularly good and mature rust implementation of Raft, that I know of indeed. I had to pick C because this lib is mainly used for replicating SQLite, which is in C, but I had also considered rust back then.

> In a second time I would like to create a pure Rust implementation of a `raft_io` backend that I will be able to include with the library. It would be easier to integrate in the async ecosystem of Rust than the `uv` based one and it would be more cross-platform too.
[...]

PR #132, opened 2020-04-20
Add raft_aligned_free() API for releasing aligned memory

freeekanayaka:
This should help ports to other platforms (e.g. Windows, #119) and language bindings (e.g. Rust, #130).

### DIFF
--- a/src/heap.c
+++ b/src/heap.c
@@ -34,13 +34,20 @@ static void *defaultAlignedAlloc(void *data, size_t alignment, size_t size)
     return aligned_alloc(alignment, size);
 }
 
+static void defaultAlignedFree(void *data, size_t alignment, void *ptr)
+{
+    (void)alignment;
+    defaultFree(data, ptr);
+}
+
 static struct raft_heap defaultHeap = {
-    NULL,               /* data */
-    defaultMalloc,      /* malloc */
-    defaultFree,        /* free */
-    defaultCalloc,      /* calloc */
-    defaultRealloc,     /* realloc */
-    defaultAlignedAlloc /* aligned_alloc */
+    NULL,                /* data */
+    defaultMalloc,       /* malloc */
+    defaultFree,         /* free */
+    defaultCalloc,       /* calloc */
+    defaultRealloc,      /* realloc */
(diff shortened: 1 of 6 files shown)


freeekanayaka:
@AshfordN, @Kerollmops does this look sane? Is it enough for helping you?

Kerollmops:
Seems good to me, I think I'll be able to make a good Rust binding to the global allocator!
Thank you!

src/heap.c
```diff
@@ -34,13 +34,20 @@ static void *defaultAlignedAlloc(void *data, size_t alignment, size_t size)
     return aligned_alloc(alignment, size);
 }
 
+static void defaultAlignedFree(void *data, size_t alignment, void *ptr)
+{
+    (void)alignment;
```
brauner, line 39: Can you add a short comment explaining why this cast to `(void)` here exists, please?

test/lib/heap.c
```diff
@@ -119,7 +131,7 @@ void test_heap_tear_down(struct raft_heap *h)
 {
     struct heap *heap = h->data;
     if (heap->n != 0) {
-      //munit_errorf("memory leak: %d outstanding allocations", heap->n);
+        // munit_errorf("memory leak: %d outstanding allocations", heap->n);
```
brauner, line 134: Is this left here on purpose?

src/heap.c
freeekanayaka: The cast to (void) is simply to silence warnings due to unused variables.

We might want to introduce something like:

```
#define RAFT_UNUSED __attribute__((__unused__))
```

and then:

```
static void defaultAlignedFree(void *data, RAFT_UNUSED size_t alignment, void *ptr)
{
```

however that'd be probably compiler-specific and might vary across platforms.

test/lib/heap.c
freeekanayaka: Yeah it's left on purpose. That assertion is currently broken, it's basically a test-level issue.

freeekanayaka: Leaks in tests are mainly detected using the memory sanitizer.

PR #133, opened 2020-04-24
Fix some coverity warnings

freeekanayaka:
Fix a few of the warnings reported by Coverity.

### DIFF
--- a/src/err.h
+++ b/src/err.h
@@ -41,12 +41,12 @@
 void errMsgWrap(char *e, const char *format);
 
 /* Transfer an error message from an object to another, wrapping it. */
-#define ErrMsgTransfer(ERRMSG1, ERRMSG2, FORMAT)     \
-    strncpy(ERRMSG2, ERRMSG1, RAFT_ERRMSG_BUF_SIZE); \
+#define ErrMsgTransfer(ERRMSG1, ERRMSG2, FORMAT)    \
+    memcpy(ERRMSG2, ERRMSG1, RAFT_ERRMSG_BUF_SIZE); \
     ErrMsgWrapf(ERRMSG2, FORMAT)
 
 #define ErrMsgTransferf(ERRMSG1, ERRMSG2, FORMAT, ...) \
-    strncpy(ERRMSG2, ERRMSG1, RAFT_ERRMSG_BUF_SIZE);   \
+    memcpy(ERRMSG2, ERRMSG1, RAFT_ERRMSG_BUF_SIZE);    \
     ErrMsgWrapf(ERRMSG2, FORMAT, __VA_ARGS__)
 
 /* Use the static error message for the error with the given code. */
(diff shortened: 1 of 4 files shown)


src/err.h
```diff
@@ -41,12 +41,12 @@
 void errMsgWrap(char *e, const char *format);
 
 /* Transfer an error message from an object to another, wrapping it. */
-#define ErrMsgTransfer(ERRMSG1, ERRMSG2, FORMAT)     \
-    strncpy(ERRMSG2, ERRMSG1, RAFT_ERRMSG_BUF_SIZE); \
+#define ErrMsgTransfer(ERRMSG1, ERRMSG2, FORMAT)    \
+    memcpy(ERRMSG2, ERRMSG1, RAFT_ERRMSG_BUF_SIZE); \
```
freeekanayaka, line 45: This might look odd, but it's safe in practice. Both ERRMSG2 and ERRMSG1 are pointers to buffers of RAFT_ERRMSG_BUF_SIZE bytes. Unfortunately Coverity can't figure that out and emits a warning.

src/err.h
```diff
@@ -41,12 +41,12 @@
...
     ErrMsgWrapf(ERRMSG2, FORMAT)
 
 #define ErrMsgTransferf(ERRMSG1, ERRMSG2, FORMAT, ...) \
-    strncpy(ERRMSG2, ERRMSG1, RAFT_ERRMSG_BUF_SIZE);   \
+    memcpy(ERRMSG2, ERRMSG1, RAFT_ERRMSG_BUF_SIZE);    \
```
freeekanayaka, line 49: Same as above.

src/err.h
freeekanayaka: One alternative might probably be using `strlcpy`, but that's not in the standard lib afaiu.

src/log.c
```diff
@@ -267,11 +267,14 @@ static void refsIncr(struct raft_log *l,
 
     /* Lookup the slot associated with the given term/index, which must have
      * been previously inserted. */
-    for (slot = &l->refs[key]; slot != NULL; slot = slot->next) {
+    slot = &l->refs[key];
+    while (1) {
```
freeekanayaka, line 271: This is safe to do because, by construction, we know that we'll find a slot with `slot->term == term` and the loop will terminate. This new version avoids comparing to `slot` to `NULL` which makes Coverity unhappy (since the code later dereferences `slot`).

src/log.c
```diff
@@ -297,14 +300,16 @@ static bool refsDecr(struct raft_log *l,
 
     /* Lookup the slot associated with the given term/index, keeping track of
      * its previous slot in the bucket list. */
-    for (slot = &l->refs[key]; slot != NULL; slot = slot->next) {
+    slot = &l->refs[key];
+    while (1) {
```
freeekanayaka, line 304: Same as above.

PR #139, opened 2020-05-21
Grant out vote to the transferee if we're currently the leader

freeekanayaka:
When transferring leadership, we should grant our vote to the target leader.


freeekanayaka:
Travis is actually done and green https://travis-ci.org/github/canonical/raft/builds/689709126.

Not sure why the build still shows up as pending here.

PR #160, opened 2020-11-16
Fix conversion to 'int' from 'unsigned int'

ericcurtin:
Fix conversion to 'int' from 'unsigned int' may change the sign of the
result warning.

Fix gcc 4.8.5 warning.

### DIFF
--- a/src/uv_os.c
+++ b/src/uv_os.c
@@ -201,7 +201,7 @@ int UvOsIoGetevents(aio_context_t ctx,
     return rv;
 }
 
-int UvOsEventfd(unsigned int initval, int flags)
+int UvOsEventfd(int initval, int flags)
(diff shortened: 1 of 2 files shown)


ericcurtin:
I understand these kind of warnings are the lowest of the low in terms of low-hanging fruit. Fixed these while looking at an issue.

freeekanayaka:
I think you might fixed a warning but created a new one, at least in the gcc/clang versions that are running in Travis. See the failed tests.

ericcurtin:
They changed the function prototype:

https://github.com/bminor/glibc/commit/69eb9a183c19e8739065e430758e4d3a2c5e4f1a

centos7 is pretty old I guess. Closing...

PR #165, opened 2021-01-23
add a simple cmake

linux-fan-dave:
This commit adds a simple cmake file. This allows the code to be easier ported to other OSs - especially if they do not support pkg-config like Visual Studio.


MathieuBordere:
I'm not really planning to maintain a cmake file for this library, will close for now.

PR #169, opened 2021-02-15
Mbordere/barrier wait for segments

MathieuBordere:
A UvBarrier's Callback should only fire when all open segments that reference the barrier have been finalized. 

Before this PR, that was not the case, the callback fired once the first open segment referencing the barrier was finalized.
This occurred when a follower was installing a snapshot; eventually `uvSnapshotPutFinish` was called, freeing the put request's memory, including the memory of the UvBarrier. Because the barrier CB fired early, there were still segments referencing the barrier and one of them would become the active barrier due to the logic after the `start` label in `uvAppendMaybeStart`. This would lead to accessing freed memory when attempting to call `uv->barrier->cb` in `uvFinalizeAfterWorkCb`.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -167,6 +167,7 @@ test_unit_uv_CFLAGS = $(AM_CFLAGS) -Wno-conversion
 test_unit_uv_LDADD = libtest.la
 
 test_integration_uv_SOURCES = \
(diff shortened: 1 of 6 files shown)


MathieuBordere:
checking the CI failure + minor aesthetic fix

PR #170, opened 2021-02-16
Mbordere/barrier wait for segments

MathieuBordere:
A UvBarrier's Callback should only fire when all open segments that reference the barrier have been finalized.

Before this PR, that was not the case, the callback fired once the first open segment referencing the barrier was finalized.
This could occur when a follower was installing a snapshot; eventually `uvSnapshotPutFinish` was called, freeing the put request's memory, including the memory of the `UvBarrier`. Because the barrier CB fired early, there were still segments referencing the barrier and one of them would become the active barrier due to the logic after the `start` label in `uvAppendMaybeStart`. This would lead to accessing freed memory when attempting to call `uv->barrier->cb` in `uvFinalizeAfterWorkCb`.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -167,6 +167,7 @@ test_unit_uv_CFLAGS = $(AM_CFLAGS) -Wno-conversion
 test_unit_uv_LDADD = libtest.la
 
 test_integration_uv_SOURCES = \
+  ${libraft_la_SOURCES} \
   test/integration/main_uv.c \
   test/integration/test_uv_init.c \
   test/integration/test_uv_append.c \
--- a/test/integration/test_uv_append.c
+++ b/test/integration/test_uv_append.c
@@ -1,6 +1,7 @@
 #include "../lib/runner.h"
 #include "../lib/uv.h"
 #include "../lib/aio.h"
+#include "../../src/uv.h"
 
 /* Maximum number of blocks a segment can have */
 #define MAX_SEGMENT_BLOCKS 4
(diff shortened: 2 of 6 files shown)


Makefile.am
```diff
@@ -167,6 +167,7 @@ test_unit_uv_CFLAGS = $(AM_CFLAGS) -Wno-conversion
 test_unit_uv_LDADD = libtest.la
 
 test_integration_uv_SOURCES = \
+  ${libraft_la_SOURCES} \
```
stgraber, line 170: Looks unrelated to the rest of the commit, should this be split out?

test/integration/test_uv_append.c
```diff
@@ -65,19 +67,23 @@ static void appendCbAssertResult(struct raft_io_append *req, int status)
         }                                                   \
     }
 
-/* Submit an append request identified by I, with N_ENTRIES entries, each one of
- * size ENTRY_SIZE. The default expectation is for the operation to succeed. A
- * custom STATUS can be set with APPEND_EXPECT. */
-#define APPEND_SUBMIT(I, N_ENTRIES, ENTRY_SIZE)                     \
+#define APPEND_SUBMIT_CB_DATA(I, N_ENTRIES, ENTRY_SIZE, CB, DATA)   \
```
stgraber, line 70: Shouldn't this one have a description comment?

stgraber (changes requested):
Just a couple of nits

Makefile.am
MathieuBordere: Those sources are needed for the calls to `UvBarrier` and `UvUnblock` in the new tests.

test/integration/test_uv_append.c
MathieuBordere: Ok

PR #173, opened 2021-02-20
Build on MacOS

rabits:
Patch to support the MacOS for go-dqlite, basically a copy of abandoned PR #119 with some cleaning and MacOS-specific changes.

Now it's ready for review - I tested it successfully with macos->macos and macos->linux connection and regular DB workload.
Tests are in the go-dqlite PR here: https://github.com/canonical/go-dqlite/pull/132

The most important change that fixed the freeze is moving `r->leader_state.change = req;` prior to call of the `clientChangeConfiguration` function (because the macos IO implementation is not async and executes directly, so quite sure even on linux it could fail in case the async executor will decide to run IO operation right after the call).

The other patches in series:
* https://github.com/canonical/go-dqlite/pull/132
* https://github.com/canonical/dqlite/pull/285
* https://github.com/canonical/dqlite/pull/288
* https://github.com/canonical/dqlite/pull/290

### DIFF
--- a/example/server.c
+++ b/example/server.c
@@ -13,6 +13,11 @@
 #define Logf(SERVER_ID, FORMAT, ...) \
     printf("%d: " FORMAT "\n", SERVER_ID, __VA_ARGS__)
 
+//apparently srandom isn't defined on mingw
+#if !defined(srandom)
+#define srandom srand
+#endif
+
 /********************************************************************
  *
  * Sample application FSM that just increases a counter.
@@ -166,8 +171,12 @@ static int ServerInit(struct Server *s,
     memset(s, 0, sizeof *s);
 
     /* Seed the random generator */
+#if defined(_WIN32)
+    srandom((unsigned)time(NULL));
+#else
     timespec_get(&now, TIME_UTC);
     srandom((unsigned)(now.tv_nsec ^ now.tv_sec));
+#endif
 
     s->loop = loop;
 
@@ -385,8 +394,10 @@ int main(int argc, char *argv[])
     dir = argv[1];
     id = (unsigned)atoi(argv[2]);
 
+#if !defined(_WIN32)
     /* Ignore SIGPIPE, see https://github.com/joyent/libuv/issues/1254 */
     signal(SIGPIPE, SIG_IGN);
+#endif
 
     /* Initialize the libuv loop. */
     rv = uv_loop_init(&loop);
--- a/src/client.c
+++ b/src/client.c
@@ -198,13 +198,14 @@ int raft_add(struct raft *r,
 
     req->cb = cb;
 
-    rv = clientChangeConfiguration(r, req, &configuration);
-    if (rv != 0) {
-        goto err_after_configuration_copy;
-    }
-
     assert(r->leader_state.change == NULL);
     r->leader_state.change = req;
+
+    rv = clientChangeConfiguration(r, req, &configuration);
+    if (rv != 0) {
+        r->leader_state.change = NULL;
+        goto err_after_configuration_copy;
+    }
 
     return 0;
 
@@ -352,13 +353,14 @@ int raft_remove(struct raft *r,
 
     req->cb = cb;
 
-    rv = clientChangeConfiguration(r, req, &configuration);
-    if (rv != 0) {
-        goto err_after_configuration_copy;
-    }
-
     assert(r->leader_state.change == NULL);
     r->leader_state.change = req;
+
+    rv = clientChangeConfiguration(r, req, &configuration);
+    if (rv != 0) {
+        r->leader_state.change = NULL;
+        goto err_after_configuration_copy;
+    }
 
     return 0;
 
--- a/src/heap.c
+++ b/src/heap.c
@@ -31,13 +31,34 @@ static void *defaultRealloc(void *data, void *ptr, size_t size)
 static void *defaultAlignedAlloc(void *data, size_t alignment, size_t size)
 {
     (void)data;
+#ifdef _WIN32
+    return _aligned_malloc(size, alignment);
+#elif defined(__APPLE__)
+    void * p1; // original block
+    void ** p2; // aligned block
+    size_t offset = alignment + sizeof(void *) - 1;
+    if ((p1 = (void *)malloc(size + offset)) == NULL)
+        return NULL;
+    p2 = (void **)(((uintptr_t)(p1) + offset) & ~(alignment - 1));
+    p2[-1] = p1;
+    return p2;
+#else
     return aligned_alloc(alignment, size);
+#endif
 }
 
 static void defaultAlignedFree(void *data, size_t alignment, void *ptr)
 {
     (void)alignment;
+#ifdef _WIN32
+    (void)ptr;
+    _aligned_free(data);
+#elif defined(__APPLE__)
+    (void)data;
+    free(((void * *)ptr)[-1]);
+#else
     defaultFree(data, ptr);
+#endif
 }
 
 static struct raft_heap defaultHeap = {
--- a/src/uv.c
+++ b/src/uv.c
@@ -534,12 +534,14 @@ static int uvBootstrap(struct raft_io *io,
     /* Write the term */
     rv = uvSetTerm(io, 1);
     if (rv != 0) {
+        ErrMsgPrintf(io->errmsg, "Unable to set UV term");
         return rv;
     }
 
     /* Create the first closed segment file, containing just one entry. */
     rv = uvSegmentCreateFirstClosed(uv, configuration);
     if (rv != 0) {
+        ErrMsgPrintf(io->errmsg, "Unable to create first closed segment");
         return rv;
     }
 
--- a/src/uv_fs.c
+++ b/src/uv_fs.c
@@ -2,7 +2,9 @@
 
 #include <stdlib.h>
 #include <string.h>
+#ifdef __linux__
 #include <sys/vfs.h>
+#endif
 #include <unistd.h>
 
 #include "assert.h"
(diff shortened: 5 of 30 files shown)


example/server.c
```diff
@@ -166,8 +171,13 @@ static int ServerInit(struct Server *s,
     memset(s, 0, sizeof *s);
 
     /* Seed the random generator */
+#if defined(_WIN32)
+    /*mingw doesn't implement timespec_get*/
+    srandom(5);
```
MathieuBordere, line 176: something like `srand((unsigned)time(NULL));` ?

src/client.c
```diff
@@ -198,14 +198,14 @@ int raft_add(struct raft *r,
 
     req->cb = cb;
 
+    assert(r->leader_state.change == NULL);
+    r->leader_state.change = req;
```
MathieuBordere, line 202: You should reset `r->leader_state.change` to NULL if `clientChangeConfiguration` fails, same for the other similar changes.

src/heap.c
```diff
@@ -31,13 +31,40 @@ static void *defaultRealloc(void *data, void *ptr, size_t size)
...
+#ifdef _WIN32
+    /*
+        Windows does not have support for aligned_alloc, the closest functionality
+        is provided by _aligned_malloc(), but the returned memory has to be freed with
+        _aligned_free(). Freeing this with free() would not properly clean up the memory.
+        TODO: This should be addressed
+    */
+    return _aligned_malloc(size, alignment);
+#elif defined(__APPLE__)
+    void * p1; // original block
```
MathieuBordere, line 43: Indentation is off in this codeblock.

src/heap.c
```diff
@@ -31,13 +31,40 @@ static void *defaultRealloc(void *data, void *ptr, size_t size)
...
+	void ** p2; // aligned block
+	size_t offset = alignment + sizeof(void *) - 1;
+	if ((p1 = (void *)malloc(size + offset)) == NULL)
+		return NULL;
+	p2 = (void **)(((size_t)(p1) + offset) & ~(alignment - 1));
+	p2[-1] = p1;
```
MathieuBordere, line 49: What does this line do?

src/uv.c
```diff
@@ -107,6 +107,7 @@ static int uvStart(struct raft_io *io,
     }
     rv = uv_timer_start(&uv->timer, uvTickTimerCb, msecs, msecs);
     assert(rv == 0);
+
```
MathieuBordere, line 110: Spurious newline :-D

src/uv_fs.c
```diff
@@ -381,7 +408,16 @@ int UvFsMakeOrOverwriteFile(const char *dir,
         goto err;
     }
 
+#ifdef WINDOWS
+    uv_buf_t convBufArr[1];
+    uv_buf_t convBuf;
```
MathieuBordere, line 413: the separate convBuf variable is not needed, you could work with `uv_buf_t` in the array directly.

src/uv_os.h
```diff
@@ -4,7 +4,9 @@
 #define UV_OS_H_
 
 #include <fcntl.h>
-#include <linux/aio_abi.h>
+#ifdef __linux__
+#    include <linux/aio_abi.h>
```
MathieuBordere, line 8: No indentation needed for me.

example/server.c
rabits: Agree, missed this thing.

src/heap.c
```diff
@@ -31,13 +31,40 @@ static void *defaultRealloc(void *data, void *ptr, size_t size)
...
 {
     (void)data;
+#ifdef _WIN32
+    /*
+        Windows does not have support for aligned_alloc, the closest functionality
+        is provided by _aligned_malloc(), but the returned memory has to be freed with
+        _aligned_free(). Freeing this with free() would not properly clean up the memory.
+        TODO: This should be addressed
+    */
+    return _aligned_malloc(size, alignment);
```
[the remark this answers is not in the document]
rabits: I think it's handled in `defaultAlignedFree`, so will remove TODO from here.

src/heap.c
rabits: Honestly I don't know - aligned malloc is way beyond my C knowledge) But it's the best answer I found: https://stackoverflow.com/a/38089003 : The line stores the original pointer before the aligned one, if I did understand it correctly.

src/uv.c
rabits: Oh I thought I cleaned all my "it need to be perfect everywhere" changes)

src/uv_fs.c
rabits: Agree, switched to "slim" version.

src/uv_os.h
rabits: Yep, fixed

src/heap.c
rabits: Fixed, oh those tabs...

src/client.c
rabits: Yep, I think it will be perfect)

src/heap.c
MathieuBordere: Hehe, yeah, it looks a bit fishy at first sight but it's only pointer arithmetic, so is indeed fine.
There's a comment in the stackoverflow thread 
```
Also: size_t (in the line that sets p2) should be uintptr_t. There's no guarantee that size_t is large enough to represent pointer values.
```
Could you add that please?

rabits: Oh, good eye! Thank you - just fixed it.

rabits:
Hi @stgraber could you please check this change as well?

stgraber:
I think I'd feel a lot more confident about this (and easier to review/revert) if it was split into a bunch of commits for the various type of changes done in there.

rabits:
@stgraber you mean half-working ones? I checked the changelist again and almost any line is needed to be sure that the project could be built or work at all... Maybe you have some clues on how to split it, because it's quite none in my head right now...

pycaster:
@rabits Thank you for the PR. I am looking forward for this to land on master. However, I got your changes to build on MAC with this patch to raft, `master` of dqlite but seeing this error when I ran the dqlite-demo example,

```
dqlite-demo --api 127.0.0.1:8001 --db 127.0.0.1:9001
Assertion failed: (amount == (int)page_size), function vfsWalWrite, file src/vfs.c, line 1079.
[1]    79045 abort      dqlite-demo --api 127.0.0.1:8001 --db 127.0.0.1:9001
```

Any idea what's going here?

rabits:
Hi @pycaster , yeah, still working on the patches - but with the latest releases. So still need to prepare CI and retest them again.

Regarding the error you see - looks like some assert failed, hard to tell right now, but let's check when CI part will be ready.

rabits:
@MathieuBordere , I tested osx build but looks like it's disabled for now - do you know a way to make it through?

rabits:
Oh, sorry - hopefully it's not me who eaten the budget... Is osx pricey on travis?

rabits:
Hi @MathieuBordere , could you please check the CI log - I'm a little bit confused why the test could fail:
```
send/changeToUnconnectedAddress                             [ ERROR ]
Error: test/integration/test_uv_send.c:281: uv_run: condition not met in 20 iterations
Error: child killed by signal 6
```

Also not sure if the other tests are fixed correctly, so am I going in the right direction?

MathieuBordere:
> Hi @MathieuBordere , could you please check the CI log - I'm a little bit confused why the test could fail:
> 
> ```
> send/changeToUnconnectedAddress                             [ ERROR ]
> Error: test/integration/test_uv_send.c:281: uv_run: condition not met in 20 iterations
> Error: child killed by signal 6
> ```
> 
> Also not sure if the other tests are fixed correctly, so am I going in the right direction?

It's not obvious why it would fail, I propose that you try to run the test locally and debug it.

steps are:
- Try to send a message to a server at a certain address
- Try to send a message to the same server at a different, disconnected address, hitting the code path https://github.com/canonical/raft/blob/34b3ae06efdb221c819ef4c804c37128008467df/src/uv_send.c#L426 but the connection will fail. 
- Try to send a message to the same server at the original, connected address (this is where it fails for you). My guess is that https://github.com/canonical/raft/blob/34b3ae06efdb221c819ef4c804c37128008467df/src/uv_send.c#L487 has a non-0 return value.

penberg:
@MathieuBordere I can reproduce the problem locally.

It does not look like `uvGetClient()` is the culprit (it can't return non-zero unless memory allocation fails AFAICT).

Can you walk me through how a new connection is supposed to appear in the test and who should flush the pending messages?

I hacked `tracing.h` as follows:

```patch
diff --git a/src/tracing.h b/src/tracing.h
index aeaddb2..9b0d9e2 100644
--- a/src/tracing.h
+++ b/src/tracing.h
@@ -16,6 +16,8 @@ extern struct raft_tracer StderrTracer;
 /* Emit a debug message with the given tracer. */
 #define Tracef(TRACER, ...)                                              \
     do {                                                                 \
+fprintf(stderr, __VA_ARGS__); \
+fprintf(stderr, "\n"); \
         if (TRACER != NULL && TRACER->emit != NULL && TRACER->enabled) { \
             static char _msg[1024];                                      \
             snprintf(_msg, sizeof _msg, __VA_ARGS__);                    \
```

and see the following:

```
penberg@vonneumann raft % make test/integration/uv && ./test/integration/uv send/changeToUnconnectedAddress
make: `test/integration/uv' is up to date.
Running test suite with seed 0xc89a4fe4...
send/changeToUnconnectedAddress                             [ ERROR ]
no connection available -> enqueue message
connect attempt completed -> status unknown error
send pending messages
connection available -> write message
no connection available -> enqueue message
no connection available -> enqueue message
Error: test/integration/test_uv_send.c:284: uv_run: condition not met in 20 iterations
Error: child killed by signal 6
0 of 1 (0%) tests successful, 0 (0%) test skipped.
```

MathieuBordere:
Thanks for this, will have a look later this week!

calvin2021y:
a lot developer use MacOS, please consider fix this.

rabits:
Oh, just recalled that this PR is still here. Sorry folks, I used dqlite in my project before, but figured out some other solution and quite sure will not going to continue work on this PR - gets too complicated to support this. Don't really mind if someone wants to continue the work)

PR #174, opened 2021-02-23
Mbordere/retry install snapshot

MathieuBordere:
Retries the InstallSnapshot RPC after a timeout. The retries also work as "HeartBeats" during a Follower's Snapshot installation keeping the follower from converting to a candidate. In a next phase I would like to send real HeartBeat AppendEntries RPCs regardless if the follower is installing a Snapshot or not, but that requires more profound changes.

In this PR I have also changed the logic when `last_send` is updated, next to being updated when sending an AppendEntries RPC it is also updated when sending an InstallSnaphot RPC. I also update it synchronously in the tickCb, otherwise we risk sending multiple RPCs while we wait for `last_send` to be updated in the respective RPC send callbacks.

### DIFF
--- a/include/raft.h
+++ b/include/raft.h
@@ -613,6 +613,12 @@ struct raft
     unsigned heartbeat_timeout;
 
     /*
+     * When the leader sends an InstallSnapshot RPC to a follower it will consider
+     * the RPC as failed after this timeout and retry.
+     */
+    unsigned install_snapshot_timeout;
+
+    /*
      * The fields below hold the part of the server's volatile state which is
      * always applicable regardless of the whether the server is follower,
      * candidate or leader (Figure 3.1). This state is rebuilt automatically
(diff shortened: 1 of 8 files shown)


freeekanayaka:
I have some feeling that updating `last_send` asynchronously had some specific reason. But I really can't recall at the moment, I might well be off. I didn't look at the diff yet, but I'd suggest to double check that part (I think it's fine, but sounds the most delicate change here).

MathieuBordere:
> I have some feeling that updating `last_send` asynchronously had some specific reason. But I really can't recall at the moment, I might well be off. I didn't look at the diff yet, but I'd suggest to double check that part (I think it's fine, but sounds the most delicate change here).

The update to `last_send` was moved to the callback in https://github.com/canonical/raft/commit/f09f9e789aa564a2ed98e6f50dc16323f0f744f0 , can't really see if there was a reason to do it, maybe seeing the code rings a bell for you?

freeekanayaka:
> > I have some feeling that updating `last_send` asynchronously had some specific reason. But I really can't recall at the moment, I might well be off. I didn't look at the diff yet, but I'd suggest to double check that part (I think it's fine, but sounds the most delicate change here).
> 
> The update to `last_send` was moved to the callback in [f09f9e7](https://github.com/canonical/raft/commit/f09f9e789aa564a2ed98e6f50dc16323f0f744f0) , can't really see if there was a reason to do it, maybe seeing the code rings a bell for you?

Unfortunately it does not ring a bell :/ If there's no failure in the test suite and no difference in real world tests maybe it doesn't matter.

PR #176, opened 2021-03-04
Mbordere/refresh connections cache

MathieuBordere:
Fixes an issue where in case of a reused `raft_id`, a connection would be attempted to a wrong address. This happened when LXD removed a server from the configuration and later added a new one to the configuration, with the same id and different address.
It could also happen when the same raft node has changed its address.

The issue is caused by `uvGetClient` returning a cached client object.

### DIFF
--- a/src/uv_send.c
+++ b/src/uv_send.c
@@ -374,13 +374,13 @@ static void uvClientTimerCloseCb(struct uv_handle_s *handle)
     uvClientMaybeDestroy(c);
 }
 
-/* Start shutting down a client since the raft_io instance has been closed. */
+/* Start shutting down a client. This happens when the `raft_io` instance
+ * has been closed or when the address of the client has changed. */
 static void uvClientAbort(struct uvClient *c)
 {
     struct uv *uv = c->uv;
     int rv;
 
-    assert(uv->closing);
     assert(c->stream != NULL || c->old_stream != NULL ||
            uv_is_active((struct uv_handle_s *)&c->timer) ||
            c->connect.data != NULL);
@@ -421,8 +421,13 @@ static int uvGetClient(struct uv *uv,
         if ((*client)->id != id) {
             continue;
         }
-        /* TODO: handle a change in the address */
-        /* assert(strcmp((*client)->address, address) == 0); */
+
+        /* Client address has changed, abort connection and create a new one. */
+        if (strcmp((*client)->address, address) != 0) {
+            uvClientAbort(*client);
+            break;
+        }
+
         return 0;
     }
 
--- a/test/integration/test_uv_send.c
+++ b/test/integration/test_uv_send.c
@@ -259,6 +259,35 @@ TEST(send, badAddress, setUp, tearDownDeps, 0, NULL)
     return MUNIT_OK;
 }
 
+/* Make sure UvSend doesn't use a stale connection for a certain server id
+ * by first sending a message to a valid address and then sending a message to
+ * an invalid address, making sure the valid connection is not reused.
+ * Afterwards assert that a send to the correct address still succeeds. */
+TEST(send, changeToUnconnectedAddress, setUp, tearDownDeps, 0, NULL)
+{
+    struct fixture *f = data;
+
+    /* Send a message to a server and a connected address */
+    SEND(0);
+
+    /* Send a message to the same server, but update the address to an
+     * unconnected address and assert it fails. */
+    munit_assert_ullong(MESSAGE(0)->server_id, ==, MESSAGE(1)->server_id);
+    MESSAGE(1)->server_address = "127.0.0.2:1";
+    SEND_SUBMIT(1 /* message */, 0 /* rv */, RAFT_CANCELED /* status */);
+
+    /* Send another message to the same server and connected address */
+    munit_assert_ullong(MESSAGE(0)->server_id, ==, MESSAGE(2)->server_id);
+    SEND(2);
+
+    /* Send another message to the same server and connected address */
+    munit_assert_ullong(MESSAGE(0)->server_id, ==, MESSAGE(3)->server_id);
+    SEND(3);
+
+    TEAR_DOWN_UV;
+    return MUNIT_OK;
+}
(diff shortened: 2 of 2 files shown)


freeekanayaka:
A server ID should not be re-used. I think there might be assumptions around that (like this one). Would it be difficult to fix the LXD side instead? To not re-use the same ID. I'm not sure how complicated that would be, if it turns out to be problematic, we might need to watch out for other places in libraft where re-using an ID might be problematic.

stgraber:
The design of the `nodes` table didn't include an AUTOINCREMENT field, so instead IDs get reused...
We can change the schema moving forward but can't do anything about 4.0 sadly...

stgraber:
@tomponline can you send a tiny fix to make it AUTOINCREMENT? I think our investigation yesterday showed that we don't have a problem with it being auto-increment, so may be good to have for 4.12.

MathieuBordere:
I believe @stgraber said it was a straightforward fix for new versions of LXD, but that a 

> @tomponline can you send a tiny fix to make it AUTOINCREMENT? I think our investigation yesterday showed that we don't have a problem with it being auto-increment, so may be good to have for 4.12.

I made a build for @morphis including AUTOINCREMENT on the nodes table, but it didn't resolve the issue, id's were being reused. But maybe I botched the build ...

tomponline:
@MathieuBordere how did you do the patch?

MathieuBordere:
> @MathieuBordere how did you do the patch?

https://github.com/MathieuBordere/lxd/commit/baeae807466c6cd56cf9b290bf1743a75551d1ba

stgraber:
right, we need an upgrade patch to mangle the existing table.

tomponline:
@stgraber this might be quite involved, because as far as I can see we cannot add autoincrement to an existing column as a direct patch, so we need to create a new `nodes` table, copy the contents, delete the old one and rename the new one in its place. However when we do that any rows in other tables referencing that node in other tables will be removed, which in turn could remove rows in other tables that also had those tables. The use of `ON DELETE CASCADE` is quite dangerous for that (as opposed to preventing the query from going again) and I've been burned by that in the past, as the patch can run cleanly and result in rows in (seemingly) unrelated tables being removed.

So I need to audit the related relationships all the way down, take copies of those tables and then reinsert their contents afterwards.

tomponline:
@stgraber I'll try it, there's also `PRAGMA foreign_keys = OFF;`which I remember trying in the past and it had no effect, but not sure if thats something to do with it being dqlite or just that the feature isn't built into our sqlite.

stgraber:
@tomponline foreign_keys can't be used within a transaction.

stgraber:
well, is a no-op when used within a transaction

tomponline:
@stgraber defer_foreign_keys doesn't appear to work either, and legacy_alter_table is required to avoid breaking references used in the views (although not the end of the world as relatively easy to delete and recreate them).

```sql
PRAGMA defer_foreign_keys = ON;
PRAGMA legacy_alter_table = ON;
CREATE TABLE nodes_new (
    id INTEGER PRIMARY KEY AUTOINCREMENT NOT NULL,
    name TEXT NOT NULL,
    description TEXT DEFAULT '',
    address TEXT NOT NULL,
    schema INTEGER NOT NULL,
    api_extensions INTEGER NOT NULL,
    heartbeat DATETIME DEFAULT CURRENT_TIMESTAMP,
    pending INTEGER NOT NULL DEFAULT 0,
    arch INTEGER NOT NULL DEFAULT 0 CHECK (arch > 0),
    failure_domain_id INTEGER DEFAULT NULL REFERENCES nodes_failure_domains (id) ON DELETE SET NULL,
    UNIQUE (name),
    UNIQUE (address)
);

INSERT INTO nodes_new (id, name, description, address, schema, api_extensions, heartbeat, pending, arch, failure_domain_id)
	SELECT id, name, description, address, schema, api_extensions, heartbeat, pending, arch, failure_domain_id FROM nodes;

DROP TABLE nodes;
ALTER TABLE nodes_new RENAME TO nodes;
```

This patch results in (at least) all entries in the storage_pools_config table being removed, rendering LXD unstartable.

tomponline:
@stgraber we may need to handle not overlapping ids in the insert query. I wonder if sqlite is like mysql, in that you can generate your own id as a sub query `insert into nodes (id) values((select max(id)+1 from nodes))` and then use that as the result for auto inc id.

stgraber:
That won't help us as that's the logic we have today. We do pick the next id in the table, it's just that if we deleted rows before, the new id may reuse a deleted one.

tomponline:
@stgraber ah I see, I guess we need a new table to store a count for this table then.

stgraber:
@tomponline that's more hackish than I'd like... We should at least make it autoincrement for all new clusters, that's pretty trivial by just editing the existing CREATE TABLE calls in update.go.

Did you check if PRAGMA foreign_keys=off; lets us do what we want when not run as part of a transaction?
If so, then maybe we can add an exception for that one DB patch not to run in a transaction.

tomponline:
https://github.com/lxc/lxd/pull/8521

tomponline:
@stgraber as the update functions are only passed a `tx *sql.Tx` ive not tried that yet. I'll look at how the DB schema update system works and see if I can get a raw DB handle from it.

stgraber:
My thought is that if it works we'd do it through a patch instead of a schema update.  That's not ideal and we usually avoid it but that would allow for it to be done on older releases too.

We'd put logic in that patch to only run on the leader so we don't run it multiple times.

tomponline:
@stgraber ah ok

tomponline:
@stgraber it does appear to work outside of a transaction:

https://github.com/lxc/lxd/pull/8523

tomponline:
@stgraber it does appear to work outside of a transaction:

https://github.com/lxc/lxd/pull/8523

stgraber:
nice!

tomponline:
Related https://github.com/lxc/lxd/pull/8521

PR #177, opened 2021-03-05
Mbordere/timeout connection attempts

MathieuBordere:
This will cleanup clients in the client queue to which we cannot connect, instead of retrying to connect to them forever. If raft would attempt to connect to such an unreachable client again, a new uvClient struct will be created and the retry cycle restarts.

This is an attempt to cleanup connections in the case where multiple servers are added and removed from the configuration over time and that cannot be reached anymore. Ideally I want to actively cleanup the connection to a server when the server is no longer part of the configuration, but that requires some more refactoring. What are your thoughts @freeekanayaka ?

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1,5 +1,5 @@
 AC_PREREQ(2.60)
-AC_INIT([raft], [0.10.0])
+AC_INIT([raft], [0.11.0])
 AC_LANG([C])
 AC_CONFIG_MACRO_DIR([m4])
 AC_CONFIG_AUX_DIR([ac])
--- a/include/raft/uv.h
+++ b/include/raft/uv.h
@@ -114,6 +114,12 @@ RAFT_API void raft_uv_set_segment_size(struct raft_io *io, size_t size);
 RAFT_API void raft_uv_set_connect_retry_delay(struct raft_io *io, unsigned msecs);
 
 /**
+ * Set how many attempts should be made to connect with another server.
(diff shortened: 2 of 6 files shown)


freeekanayaka:
What would happen if a server went down for a long time (e.g. maintenance) and then back online? Would a connection eventually be retried?

MathieuBordere:
> What would happen if a server went down for a long time (e.g. maintenance) and then back online? Would a connection eventually be retried?

If a follower goes down for a long time, then the leader will retry sending `AppendEntries` RPC's according to the heartbeat timeout and a connection will be setup again when `UvSend` is called. If a leader goes down for a long time, the follower's election timeout will trigger and a connection will be set up again in response to sending `RequestVote` RPC's and then later again with `AppendEntries` RPCs when the follower would convert to leader. Am i missing something?

MathieuBordere:
@freeekanayaka Do you have any objections on merging this?

configure.ac
```diff
@@ -1,5 +1,5 @@
 AC_PREREQ(2.60)
-AC_INIT([raft], [0.10.0])
+AC_INIT([raft], [0.11.0])
```
stgraber, line 2: commit message includes `wq` ;)

MathieuBordere:
Doesn't really need to be merged for me, it would address a potential small resource leak in case servers stick around that are never reachable anymore. Will close for now.

PR #178, opened 2021-03-15
release: 0.10.0

MathieuBordere:
Forgot to bump the library minor version when extending the API.

- Add raft_set_install_snapshot_timeout to API

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1,5 +1,5 @@
 AC_PREREQ(2.60)
-AC_INIT([raft], [0.9.24])
+AC_INIT([raft], [0.10.0])
 AC_LANG([C])
 AC_CONFIG_MACRO_DIR([m4])
 AC_CONFIG_AUX_DIR([ac])


stgraber:
@freeekanayaka what was your plan on libraft versioning?

freeekanayaka:
> @freeekanayaka what was your plan on libraft versioning?

I wanted to redesign a bit the public API, and have a cleaner (at least imo) control flow then the current callback based design. At that point I'd switch version to 1.0 and call it stable. I had made a lot of progress towards that design here: https://github.com/freeekanayaka/raft/tree/decouple-from-uv. There's a bit of background around the rational for the redesign in [this](https://github.com/canonical/raft/issues/130#issuecomment-615321327) comment and following ones.

All that being said, I don't think that design will ever come to light any time soon. So if there's a feeling that we should tag a 1.0, I think that's fine. The new design could be 2.0, if it ever realizes.

stgraber:
@freeekanayaka this PR bumps us from 0.9.24 to 0.10.0.
I was basically wondering if this should be 0.9.25 instead or if there's a good reason to make it 0.10.0.

I'd probably hold off until we have more extensive CI and jepsen testing before calling anything 1.0 for sure.

freeekanayaka:
> @freeekanayaka this PR bumps us from 0.9.24 to 0.10.0.
> I was basically wondering if this should be 0.9.25 instead or if there's a good reason to make it 0.10.0.
> 
> I'd probably hold off until we have more extensive CI and jepsen testing before calling anything 1.0 for sure.

Since there's a new public API, 0.10.0 would probably make sense. A 0.9.25 would be a bug fix release. Both dqlite and raft have been following [semver](https://semver.org/) in that regard.

freeekanayaka:
> @freeekanayaka this PR bumps us from 0.9.24 to 0.10.0.
> I was basically wondering if this should be 0.9.25 instead or if there's a good reason to make it 0.10.0.
> 
> I'd probably hold off until we have more extensive CI and jepsen testing before calling anything 1.0 for sure.

Since there's a new public API, 0.10.0 would probably make sense. A 0.9.25 would be a bug fix release. Both dqlite and raft have been following [semver](https://semver.org/) in that regard.

tomponline:
@stgraber @freeekanayaka @MathieuBordere there appears to be a 0.9.25 release already (https://github.com/canonical/raft/releases/tag/v0.9.25) and this new release doesn't yet have a tag. Is that expected?

stgraber:
@tomponline ah, I guess I need to push a tag to line this up.
No release announcement planned as it's just a minor bump due to a new symbol being exported.

PR #181, opened 2021-03-18
Queue when busy

MathieuBordere:


MathieuBordere:
@freeekanayaka This is an attempt to queue messages when the raft instance is busy. This would be a start to solve the issue described in https://github.com/canonical/dqlite/issues/284

There are 2 issues at the moment I see with this approach.
- the tick only fires at the rate of the `heartbeat_timeout` so the queued messages are only delivered at the earliest after a heartbeat_timeout.
- Every time an attempt is made to deliver a message from the queue, the election timeout will be reset, while in fact it should be reset when the message actually arrives from the other server.

Do you have other approaches in mind that could be worthwhile to investigate?

freeekanayaka:
How hard would it be to flush the queue as soon as the snapshot has been taken? That way you would not need to wait for `heartbeat_timeout`. Also I'm wondering what it would take to make the operation of taking a snapshot compatible with receiving AppendEntries concurrently. As far as the FSM is concerned taking a snapshot should be a synchronous atomic operation, so the only asynchronous part is persisting the snapshot to disk, that is probably the tricky part to deal with if AppendEntries are being received concurrently.

All that being said, I think that the approach in this branch might be fine as a quick solution. Regarding the election timeout, why can't it be reset right away at the time the message is received instead of waiting for the queue to be flushed?

MathieuBordere:
> How hard would it be to flush the queue as soon as the snapshot has been taken? That way you would not need to wait for `heartbeat_timeout`. Also I'm wondering what it would take to make the operation of taking a snapshot compatible with receiving AppendEntries concurrently. As far as the FSM is concerned taking a snapshot should be a synchronous atomic operation, so the only asynchronous part is persisting the snapshot to disk, that is probably the tricky part to deal with if AppendEntries are being received concurrently.
> 
> All that being said, I think that the approach in this branch might be fine as a quick solution. Regarding the election timeout, why can't it be reset right away at the time the message is received instead of waiting for the queue to be flushed?

Indeed, the best solution is probably to keep receiving AppendEntries while persisting the snapshot, because even with the queueing approach we still need to wait a while to start handling AppendEntries again. Will investigate what's possible there. Wrt the election timeout, it should be possible to reset right away, just didn't want to continue working on this approach if it proved to be a dead end.

freeekanayaka:
> > How hard would it be to flush the queue as soon as the snapshot has been taken? That way you would not need to wait for `heartbeat_timeout`. Also I'm wondering what it would take to make the operation of taking a snapshot compatible with receiving AppendEntries concurrently. As far as the FSM is concerned taking a snapshot should be a synchronous atomic operation, so the only asynchronous part is persisting the snapshot to disk, that is probably the tricky part to deal with if AppendEntries are being received concurrently.
> > All that being said, I think that the approach in this branch might be fine as a quick solution. Regarding the election timeout, why can't it be reset right away at the time the message is received instead of waiting for the queue to be flushed?
> 
> Indeed, the best solution is probably to keep receiving AppendEntries while persisting the snapshot, because even with the queueing approach we still need to wait a while to start handling AppendEntries again. Will investigate what's possible there. Wrt the election timeout, it should be possible to reset right away, just didn't want to continue working on this approach if it proved to be a dead end.

I don't think it's a dead end, it feels something cheap and quick if you want to improve the situation right away. A I said I believe the complexity of concurrent TakeSnapshot and AppendEntries operations will be entirely in the `raft_io` implementation, where there are probably several assumptions about entries not being appended while the snapshot is written to disk.

Issue #179, opened 2021-03-15
Keep sending HeartBeats (empty AppendEntries RPC) after sending snapshot install


MathieuBordere:
Currently, the `install_snapshot_timeout` acts as a sort-of heartbeat [1], but this should be replaced by sending real AppendEntries RPC's instead of sending SnapshotInstall RPC's (too heavy).

[1] https://github.com/canonical/raft/blob/1685cf4ea47a697000337717a96a6b75cfff8c78/include/raft.h#L619

PR #182, opened 2021-03-19
Heartbeats during snapshot install

MathieuBordere:
This would resolve https://github.com/canonical/raft/issues/179 . Currently SnapshotInstall RPCs are abused as a sort of heartbeat during follower snapshot installation, but this is unwanted due to the potentially very large size of the SnapshotInstall RPC.


include/raft.h
```diff
@@ -508,12 +508,13 @@ enum { RAFT_UNAVAILABLE, RAFT_FOLLOWER, RAFT_CANDIDATE, RAFT_LEADER };
  */
 struct raft_progress
```
stgraber, line 509: Just confirming (sorry, not familiar with code), but this struct is in-memory only and never persisted to disk or exchange over the network, right?

Always triggered by addition of struct members in the middle of a struct ;)

MathieuBordere: Yes, it's in-memory only.

MathieuBordere:
@freeekanayaka Any feedback on this?

freeekanayaka:
Seems a fine approach to me.

Issue #183, opened 2021-03-19
UvList result can include incomplete Snapshots


MathieuBordere:
`UvList` being called in `uvSnapshotGetWorkCb` can return incomplete Snapshots, this happens because it can be called while a Snapshot is being written to disk. This can result in sending an incomplete snapshot to a follower that will crash due to trying to install the incomplete snapshot.

PR #184, opened 2021-03-23
Dont load partial snapshots

MathieuBordere:
This would resolve https://github.com/canonical/raft/issues/183. A follower can crash if it receives an incomplete snapshot from a leader. The tmp-file approach in this PR tries to ensure only fully written snapshot are ever read and sent to a follower.

### DIFF
--- a/src/uv_fs.c
+++ b/src/uv_fs.c
@@ -10,7 +10,8 @@
 #include "heap.h"
 #include "uv_os.h"
 
-#define TMP_FILE_FMT "tmp-%s"
+#define TMP_FILE_PREFIX "tmp-"
+#define TMP_FILE_FMT TMP_FILE_PREFIX "%s"
 
 int UvFsCheckDir(const char *dir, char *errmsg)
 {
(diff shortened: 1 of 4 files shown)


src/uv_fs.c
```diff
@@ -264,10 +267,87 @@ int UvFsMakeFile(const char *dir,
...
+        rv = -1;
+        goto err_after_tmp_create;
+    }
+
+    /* Rename the temp file. Remark that there is a race between the existence
+     * check and the rename, there is no `renameat2` equivalent in libuv.
+     * However, in the current implementation this should pose no problems.*/
+    UvOsJoin(dir, tmp_filename, tmp_path);
+    UvOsJoin(dir, filename, path);
+    rv = UvOsRename(tmp_path, path);
```
freeekanayaka, line 304: Perhaps it's worth calling `UvFsSyncDir` too, to ensure the directory metadata is written on disk. YMMV though

freeekanayaka (approved):
Looks good to me.

PR #191, opened 2021-04-06
Bugfixes

MathieuBordere:
- There was an issue with sending AppendEntries RPC's containing entries while a snapshot was being installed by a follower. Now, we will always send empty AppendEntries RPC's to a follower that is installing a snapshot. This should fix most of @morphis his scaling issues together with https://github.com/lxc/lxd/pull/8629. 

- Protect against an InstallSnapshot RPC that arrives while entries are being saved to disk and could lead to inconsistent state.

- Don't commit entries when handling a HeartBeat while we are installing a snapshot.

### DIFF
--- a/src/replication.c
+++ b/src/replication.c
@@ -339,11 +339,13 @@ int replicationProgress(struct raft *r, unsigned i)
             assert(prev_index < snapshot_index);
             tracef("missing entry at index %lld -> send snapshot", prev_index);
             goto send_snapshot;
-        } else if (prev_term == 0 && progress_state_is_snapshot) {
-            /* The entry is not in the log, but the peer is installing a
-             * snapshot, send an empty AppendEntries RPC */
-            prev_index = 0;
         }
+    }
+
+    /* Send empty AppendEntries RPC when installing a snaphot */
+    if (progress_state_is_snapshot) {
+        prev_index = logLastIndex(&r->log);
+        prev_term = logLastTerm(&r->log);
     }
 
     return sendAppendEntries(r, i, prev_index, prev_term);
--- a/src/replication.c
+++ b/src/replication.c
@@ -1074,7 +1074,7 @@ int replicationAppend(struct raft *r,
      */
     if (n == 0) {
         if (args->leader_commit > r->commit_index) {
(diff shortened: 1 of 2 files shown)


stgraber:
Looks reasonable to me, will give a bit of time for @freeekanayaka to take a look though as I'm definitely not familiar enough with this code ;)

freeekanayaka:
How hard would it be to have unit tests that cover one on more of the mentioned edge cases? (i.e. send AppendEntries RPCs while installing a snapshot, ignore snapshots while persisting entries, and skip commit when installing a snapshot).

MathieuBordere:
> How hard would it be to have unit tests that cover one on more of the mentioned edge cases? (i.e. send AppendEntries RPCs while installing a snapshot, ignore snapshots while persisting entries, and skip commit when installing a snapshot).

You're right, I should test them.

freeekanayaka:
> > How hard would it be to have unit tests that cover one on more of the mentioned edge cases? (i.e. send AppendEntries RPCs while installing a snapshot, ignore snapshots while persisting entries, and skip commit when installing a snapshot).
> 
> You're right, I should test them.

I'm not sure how hard it will be to reproduce those scenarios, hopefully the tooling under `test/integration` are enough. If it proves to be too much effort, it might not be worth. YMMV

MathieuBordere:
> How hard would it be to have unit tests that cover one on more of the mentioned edge cases? (i.e. send AppendEntries RPCs while installing a snapshot, ignore snapshots while persisting entries, and skip commit when installing a snapshot).

- AppendEntries RPCs while installing a snapshot: This is already covered in a test, however I have a hard time reproducing in a unit test, the issue I encountered while running a reproducer script for an Anbox scaling problem.
- ignore snapshots while persisting entries: Added a test with an InstallSnapshot RCP arriving during persisting entries
- skip commit when installing a snapshot: This is already covered, tests started failing when I changed 
  `r->commit_index = min(args->leader_commit, logLastIndex(&r->log));` to 
  `r->commit_index = min(args->leader_commit, r->last_stored);` and are fixed by adding the check to skip committing while installing a snapshot.

freeekanayaka (approved):
Looks good to me.

PR #193, opened 2021-04-09
uv_snapshot: Don't remove .meta and snapshot files in UvList

MathieuBordere:
`UvList` calls `UvSnapshotInfoAppendIfMatch` and can run concurrently
on the threadpool with `uvSnapshotPutCb` through a call from `uvSnapshotGetCb`.
This can lead to the `.meta` file being removed for a snapshot that is
not fully created, resulting in an 'orphaned' snapshot file.


MathieuBordere:
In a followup PR I will cleanup these orphaned snapshot files together with some additional data integrity checks on the segment and snapshot files.

MathieuBordere:
Fixes the cause of https://github.com/canonical/dqlite/issues/289

PR #194, opened 2021-04-12
uv_fs: Remove orphaned snapshot files on init

MathieuBordere:
Fixes https://github.com/canonical/dqlite/issues/289.

Snapshots could be orphaned in certain cases, leading to high disk usage. This is fixed here https://github.com/canonical/raft/pull/193.

This PR will clean up, at init, leftover `snapshot` files that don't have a matching `.meta` file at startup and will also cleanup `.meta` files that don't have a matching `snapshot` file.

### DIFF
--- a/src/uv.c
+++ b/src/uv.c
@@ -31,6 +31,53 @@
  * TODO: implement an exponential backoff instead.  */
 #define CONNECT_RETRY_DELAY 1000
 
+/* Cleans up files that are no longer used by the system */
+static int uvMaintenance(const char *dir, char* errmsg)
+{
+    struct uv_fs_s req;
+    struct uv_dirent_s entry;
+    int n;
+    int i;
+    int rv;
+    int rv2;
+
+    n = uv_fs_scandir(NULL, &req, dir, 0, NULL);
+    if (n < 0) {
+        ErrMsgPrintf(errmsg, "scan data directory: %s", uv_strerror(n));
+        return RAFT_IOERR;
+    }
+
+    rv = 0;
+    for (i = 0; i < n; i++) {
+        const char *filename;
+        rv = uv_fs_scandir_next(&req, &entry);
+        assert(rv == 0); /* Can't fail in libuv */
+
+        filename = entry.name;
+        /* Remove leftover tmp-files */
+        if (strncmp(filename, TMP_FILE_PREFIX, strlen(TMP_FILE_PREFIX)) == 0) {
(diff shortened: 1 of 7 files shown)


freeekanayaka:
Looks good.

Just a code structuring  nitpick: `uv_fs.c` was meant to be a generic collection of low(er)-level fs-related utilities, without any direct knowledge of higher-level things like snapshot files (and that's the reason why the unit tests only include some of the source files, they just test the lower-level stuff and leave the rest to integration tests, which should use only the public `raft_io` API so they won't change much if internal implementation details change). That helps a bit in reducing the coupling between the various parts and reasoning about the code. Including `uv.h` into `uv_fs.c` means that now there's more going on at that layer. 

I'd suggest to not include `uv.h` into `uv_fs.c` and move the new logic either in `uv.c` or `uv_snapshot.c` (or a mix of both), and to add integration tests to `test_uv_load.c` instead (which already has  `manySnapshots` test which exercises removing orphaned .meta files). That reduces the surface of private APIs for which it makes sense to have low-level unit-testing, and also tests that the maintenance logic is correctly wired/plugged and works end-to-end (should that logic need changing or refactoring the test will not need to be modified).

This are just ideas, not blocking, YMMV.

freeekanayaka (approved):

MathieuBordere:
> Looks good.
> 
> Just a code structuring nitpick: `uv_fs.c` was meant to be a generic collection of low(er)-level fs-related utilities, without any direct knowledge of higher-level things like snapshot files (and that's the reason why the unit tests only include some of the source files, they just test the lower-level stuff and leave the rest to integration tests, which should use only the public `raft_io` API so they won't change much if internal implementation details change). That helps a bit in reducing the coupling between the various parts and reasoning about the code. Including `uv.h` into `uv_fs.c` means that now there's more going on at that layer.
> 
> I'd suggest to not include `uv.h` into `uv_fs.c` and move the new logic either in `uv.c` or `uv_snapshot.c` (or a mix of both), and to add integration tests to `test_uv_load.c` instead (which already has the `manySnapshots` test which exercises removing orphaned .meta files). That reduces the surface of private APIs for which it makes sense to have low-level unit-testing, and also tests that the maintenance logic is correctly wired/plugged and works end-to-end (should that logic need changing or refactoring the test will not need to be modified).
> 
> These are just ideas, not blocking, YMMV.

makes sense, ty.

PR #206, opened 2021-05-18
Compress snapshots

MathieuBordere:
This PR introduces compression for UV snapshots and adds a CRC checksum when snapshots are compressed.

Nodes running an older raft and dqlite library are still compatible with nodes running this version of the raft library because the snapshot format has not changed and is sent over the wire in the same format as before.

Downgrading from a version with compressed snapshots to a version that does not support compressed snapshots involves decompressing the snapshot files in the data folder. Afterwards the snapshots will be usable by the older library.

Currently configuration will fail if `liblz4` is not found and the `--disable-lz4` flag is not provided to `configure`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,7 @@ addons:
     - lcov
     - linux-libc-dev
     - libuv1-dev
+    - liblz4-dev
     - btrfs-progs
     - xfsprogs
     - zfsutils-linux
--- a/Makefile.am
+++ b/Makefile.am
@@ -12,6 +12,7 @@ libraft_la_LDFLAGS = -version-info 0:7:0
 libraft_la_SOURCES = \
   src/byte.c \
   src/client.c \
+  src/compress.c \
   src/configuration.c \
   src/convert.c \
   src/election.c \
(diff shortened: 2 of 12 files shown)


MathieuBordere:
I have currently left the PR in `Draft` state as I would probably ask @ktsakalozos to make a special build for @cjohnston1158 to evaluate the change (if that is practical to do) after the review is OK.

MathieuBordere:
update: I just added a test compressing a 3GB file that segfaults and while refactoring the `configure.ac` file I broke the `--disable-lz4` option ... will fix it.

MathieuBordere:
update: `--disable-lz4` now works as expected and tests are passing now (Travis seems to be acting up though when looking at the `check` tab in the PR). However, I'm not extremely happy with the fix. This line fixes it, https://github.com/MathieuBordere/raft/blob/c4f05fd8ebf43f1509ce540373e100e2b337a63c/src/compress.c#L37 it circumvents a crash in the lz4 library on `bionic` during decompression when the blockmode is `LZ4F_blockLinked`, the default. I'm not sure what I could be doing wrong to trigger that segfault, need to investigate further.

PR #218, opened 2021-07-02
uv: Set start_index correctly when no closed segments present

MathieuBordere:
This issue came up while running the jepsen tests, steps are roughly:
- Node receives snapshot install and deletes all closed segments
preceding the snapshot.
- After installing the snapshot, the node receives AppendEntries and
adds the entries to an open segment.
- Node gets killed.
- Node boots and UvLoad closes the open segment using start_index 1,
ignoring the index of the snapshot that is present in the data dir.
- Node fails to start because the newly closed segment and the snapshot are
inconsistent with a message
`raft_start(): io: last entry on disk has index 383, which is behind
last snapshot's index 3072"`

### DIFF
--- a/src/uv.c
+++ b/src/uv.c
@@ -391,7 +391,7 @@ static int uvLoadSnapshotAndEntries(struct uv *uv,
         }
         if (segments != NULL) {
             if (segments[0].is_open) {
-                *start_index = 1;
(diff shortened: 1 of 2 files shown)


freeekanayaka (approved):
Nice catch.

PR #225, opened 2021-08-11
Debugtrace

MathieuBordere:
First stab at adding tracing to raft when `LIBRAFT_TRACE` is set on startup. Will probably have to refine the messages a bit and also wire up the libuv tracing, but having these messages would already be helpful when debugging some cases.

### DIFF
--- a/src/tracing.h
+++ b/src/tracing.h
@@ -3,17 +3,27 @@
 #ifndef TRACING_H_
 #define TRACING_H_
 
+/* If an env var with this name is found, tracing can be enabled */
+#define LIBRAFT_TRACE "LIBRAFT_TRACE"
+
 #include "../include/raft.h"
 
-/* Default no-op tracer. */
 extern struct raft_tracer NoopTracer;
 
+/* Default stderr tracer. */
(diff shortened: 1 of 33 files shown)


src/tracing.h
```diff
@@ -3,17 +3,27 @@
...
+    do {                                                                 \
+        if (TRACER != NULL && TRACER->emit != NULL && TRACER->enabled) { \
+            static char _msg[1024];                                      \
+            snprintf(_msg, sizeof _msg, __VA_ARGS__);                    \
+            TRACER->emit(TRACER, __FILE__, __LINE__, _msg);              \
+        }                                                                \
     } while (0)
 
+/* Enable the tracer if the env variable is set or disable the tracer */
+void raft_tracer_maybe_enable(struct raft_tracer *tracer, bool enabled);
```
freeekanayaka, line 27: Nitpick: iiuc this is an internal function/symbol, so it should probably be named something like "TraceMaybeEnable" or similar (convention is CamelCase for internal symbols and snail_case for public ones).

PR #231, opened 2021-09-30
Remove leader

MathieuBordere:
Fixes https://github.com/canonical/dqlite/issues/324

This was the trace, a `raft_barrier` came in after the leader with id 1 was removed from the configuration.
I think a leader shouldn't be allowed to handle requests anymore once it's removed from the configuration @freeekanayaka what do you think?
  
```
LIBRAFT src/client.c:337 remove server: id 1
LIBRAFT src/replication.c:98 send 1 entries starting at 31 to server 2 (last index 32)
LIBDQLITE 1633012650556145281 gateway__handle:991 gateway handle
LIBDQLITE 1633012650556164550 handle_query_sql:537 handle query sql
LIBDQLITE 1633012650556353278 leader__barrier:473 leader barrier
LIBRAFT src/client.c:87 barrier starting at 33
LIBRAFT src/replication.c:98 send 1 entries starting at 32 to server 2 (last index 33)
LIBRAFT src/replication.c:472 leader: written 1 entries starting at 32: status 0
LIBRAFT src/replication.c:472 leader: written 1 entries starting at 33: status 0
lxd: src/replication.c:516: appendLeaderCb: Assertion `entry->type == RAFT_CHANGE' failed.
```
Another change introduced in this PR is that upon self-election, the newly, self-elected leader of a single voter cluster applies and commits all its stored log entries.

I think the PR is an improvement of the current situation, I'm just a bit worried of corner cases where a leader removes itself from the config, somehow can't replicate the config to the other voters, can't accept any commands anymore because it has been removed from the config but keeps sending HeartBeats resulting in the other voters not starting a new election.
I wonder if's not easier/cleaner/safer to just disallow a leader from removing itself from the configuration.

### DIFF
--- a/src/client.c
+++ b/src/client.c
@@ -12,6 +12,15 @@
 
 #define tracef(...) Tracef(r->tracer, __VA_ARGS__)
 
+static int canAcceptCommand(struct raft *r) {
+    const struct raft_server *server = configurationGet(&r->configuration, r->id);
+    bool in_config = (server != NULL);
+    if (r->state != RAFT_LEADER || r->transfer != NULL || !in_config) {
+        return RAFT_NOTLEADER;
+    }
+    return 0;
+}
+
 int raft_apply(struct raft *r,
                struct raft_apply *req,
                const struct raft_buffer bufs[],
@@ -21,14 +30,16 @@ int raft_apply(struct raft *r,
     raft_index index;
     int rv;
 
+    tracef("raft_apply");
+
     assert(r != NULL);
     assert(bufs != NULL);
     assert(n > 0);
 
-    if (r->state != RAFT_LEADER || r->transfer != NULL) {
-        rv = RAFT_NOTLEADER;
+    rv = canAcceptCommand(r);
+    if (rv != 0) {
         ErrMsgFromCode(r->errmsg, rv);
-        tracef("raft_apply not leader");
+        tracef("raft_apply not leader or removed rv:%d", rv);
         goto err;
     }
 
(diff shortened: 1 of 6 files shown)


MathieuBordere:
Converted to draft to wait for some comments before merging.

freeekanayaka:
@MathieuBordere did you have a look at section 4.2.2 of the Raft dissertation yet?

One relevant quote:

```
f the existing leader is asked to remove itself from the cluster, it must step down at some point. One
straightforward approach is to use the leadership transfer extension described in Chapter 3: a leader
that is asked to remove itself would transfer its leadership to another server, which would then carry
out the membership change normally.
```

freeekanayaka:
The approach currently implemented in master should still be safe, according to the argument in that section, e.g.:

```
This approach leads to two implications about decision-making that are not particularly harmful
but may be surprising. First, there will be a period of time (while it is committing Cnew) when a
leader can manage a cluster that does not include itself; it replicates log entries but does not count
itself in majorities. Second, a server that is not part of its own latest configuration should still start
new elections, as it might still be needed until the Cnew entry is committed (as in Figure 4.6). It does
not count its own vote in elections unless it is part of its latest configuration.
```

My understanding is that the current code **should** be safe, although you could implement the leadership-transfer approach described above as a variation. Not sure where the issue you have seen comes from, maybe some subtle bug in the current implementation? I.d. the implementation differs from what is described in 4.2.2.

MathieuBordere:
@freeekanayaka Thanks for the insight, I'll probably opt for the transfer approach then and read the relevant sections in the raft dissertation.

PR #238, opened 2021-11-03
Setup package builds

stgraber:


stgraber:
@MathieuBordere this will generate source package builds every time something is merged in master. I'll then have a script fetch the artifacts and automatically sign and upload those to https://launchpad.net/~dqlite/+archive/ubuntu/dev/+packages

Issue #196, opened 2021-04-27
Why need to reset vote during pre-vote 


stone-zhang-22:
In the pre-vote process, the vote is set to 0. At the same time, there is a test case(name is preVoteWithcandidateCrash) for this scenario. I don’t understand why need to reset vote. Can someone answer it, please?

freeekanayaka:
I'm not sure to understand this question. I assume you are familiar with the pre-vote algorithm as described in the Raft dissertation. The pre-vote is a "transient" state, so it must be reset if the election doesn't actual get to the vote phase. Can you please point the exact code which you don't understand?

stone-zhang-22:
int electionStart(struct raft *r)
{
    raft_term term;
    size_t n_voters;
    size_t voting_index;
    size_t i;
    int rv;
    assert(r->state == RAFT_CANDIDATE);

    n_voters = configurationVoterCount(&r->configuration);
    voting_index = configurationIndexOfVoter(&r->configuration, r->id);

    /* This function should not be invoked if we are not a voting server, hence
     * voting_index must be lower than the number of servers in the
     * configuration (meaning that we are a voting server). */
    assert(voting_index < r->configuration.n);

    /* Sanity check that configurationVoterCount and configurationIndexOfVoter
     * have returned something that makes sense. */
    assert(n_voters <= r->configuration.n);
    assert(voting_index < n_voters);

    /* During pre-vote we don't actually increment term or persist vote, however
     * we reset any vote that we previously granted since we have timed out and
     * that vote is no longer valid. */
    if (r->candidate_state.in_pre_vote) {
        **_/* Reset vote */
        rv = r->io->set_vote(r->io, 0);
        if (rv != 0) {
            goto err;
        }
        /* Update our cache too. */
        r->voted_for = 0;_**
    } else {
        /* Increment current term */
        term = r->current_term + 1;
        rv = r->io->set_term(r->io, term);
        if (rv != 0) {
            goto err;
        }

        /* Vote for self */
        rv = r->io->set_vote(r->io, r->id);
        if (rv != 0) {
            goto err;
        }

        /* Update our cache too. */
        r->current_term = term;
        r->voted_for = r->id;
    }

    /* Reset election timer. */
    electionResetTimer(r);

    assert(r->candidate_state.votes != NULL);

    /* Initialize the votes array and send vote requests. */
    for (i = 0; i < n_voters; i++) {
        if (i == voting_index) {
[...]

PR #247, opened 2021-11-19
Pre vote bugfixes

MathieuBordere:
```bash
./test/fuzzy/core liveness/networkDisconnect --seed 0x481a5fb7
```
was a reproducer for the issue described in https://github.com/canonical/raft/issues/196

Also contains another enhancement in https://github.com/canonical/raft/pull/247/commits/ac53ca37ebbe85743424fb8ce91fd97ec57376a7 that cleans up the logic a bit.

### DIFF
--- a/src/convert.c
+++ b/src/convert.c
@@ -156,7 +156,7 @@ int convertToCandidate(struct raft *r, bool disrupt_leader)
         return RAFT_NOMEM;
     }
     r->candidate_state.disrupt_leader = disrupt_leader;
-    r->candidate_state.in_pre_vote = r->pre_vote;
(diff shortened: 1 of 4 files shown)


freeekanayaka (approved):
Looks good, I guess the fact that you had a reproducer and that it now doesn't fail anymore is enough of a validation.

MathieuBordere:
> Looks good, I guess the fact that you had a reproducer and that it now doesn't fail anymore is enough of a validation.

I will formulate a response for https://github.com/canonical/raft/issues/196#issuecomment-974198143, but my brain is a bit too fried right now to do it. This PR already addresses a part of the problem, the issue should not be closed yet until its bulletproof.

PR #255, opened 2021-11-25
Flaky io log fix

MathieuBordere:


MathieuBordere:
@freeekanayaka I'm not entirely sure how to proceed, there's this user with the problem as described in the PR description. I think the general approach to truncating the in-memory and on-disk logs is OK for the test `raft_io` implementation. 
However, in our `raft_io` uv implementation, io->truncate will only be performed on closed segments, but in this case the closed segment will most likely be corrupt because it's created from an open segment that wasn't written to properly (indicated by the non-0 status in `AppendFollowerCb`), and the truncate operation will not be executed.

Should I adapt the implementation to try and truncate the corrupt segment anyway until `raft->last_stored`, assuming the write until `raft->last_stored` occurred without errors?
Should we just shut down upon seeing disk I/O errors and let `uvLoad` take care of removing corrupted open segments upon startup?
Other ideas?

freeekanayaka:
I didn't look too close at the situation, but yeah I recall that generic handling of I/O errors that are not "disk full" might be tricky. Since we expect non-"disk full" errors to be so rare (and probably severe) I think that the approach of "stopping the line" by shutting everything down immediately, and possibly let `uvLoad` see if there's something that can be done (e.g. the error was transient) is a good first approach.

Later down the road more sophisticated strategies might be applied, but I'd wait to see the details of the I/O errors happening in real world.

PS: I'm not sure how well the "disk full" errors are handled these days, but that's probably still an area of improvement?

MathieuBordere:
> I didn't look too close at the situation, but yeah I recall that generic handling of I/O errors that are not "disk full" might be tricky. Since we expect non-"disk full" errors to be so rare (and probably severe) I think that the approach of "stopping the line" by shutting everything down immediately, and possibly let `uvLoad` see if there's something that can be done (e.g. the error was transient) is a good first approach.
> 
> Later down the road more sophisticated strategies might be applied, but I'd wait to see the details of the I/O errors happening in real world.
> 
> PS: I'm not sure how well the "disk full" errors are handled these days, but that's probably still an area of improvement?

Yeah, still need to take a closer look at handling disk-full situations.

PR #264, opened 2022-02-11
uv: barrier before every snapshot

MathieuBordere:
This solves a bug that came up during jepsen tests, it occurs when:
- a server is killed ungracefully (leading to a written open segment in
the directory when the system restarts)
- a snapshot file exists
- no closed segments exist

Because open segments are not closed before writing a snapshot, raft,
when starting up and reading the files in the data directory,
erroneously assumes that all the entries in the open segment are newer
than the entries in the snapshot, while in reality the entries are
already contained in the snapshot, leading to a wrong state.

Closing the current open segments before writing the snapshot ensures
that no old entries can mistakenly be identified as new entries, all
entries in the open segments will be newer than the snapshot.

To close the open segments before making a snapshot, we perform a `barrier`
request, but we need to make a distinction between a 'blocking' and a 'non-blocking'
barrier in order to save performance.
Both barriers close all current open segments before firing the barrier
callback, but a 'non-blocking' barrier allows writes to go through to newly
created open-segments. This non-blocking barrier is used when raft is creating
a snapshot  during regular operation, the blocking barrier is used when installing
snapshots and truncating the log.

### DIFF
--- a/src/uv_snapshot.c
+++ b/src/uv_snapshot.c
@@ -573,14 +573,10 @@ static void uvSnapshotPutAfterWorkCb(uv_work_t *work, int status)
 {
     struct uvSnapshotPut *put = work->data;
     struct uv *uv = put->uv;
-    bool is_install = put->trailing == 0;
     assert(status == 0);
     uv->snapshot_put_work.data = NULL;
     uvSnapshotPutFinish(put);
-    if (is_install) {
-        UvUnblock(uv);
-    }
-    uvMaybeFireCloseCb(uv);
+    UvUnblock(uv);
 }
 
 /* Start processing the given put request. */
@@ -612,7 +608,7 @@ static void uvSnapshotPutBarrierCb(struct UvBarrier *barrier)
     }
 
     struct uv *uv = put->uv;
-    assert(put->trailing == 0);
+    assert((barrier->blocking && put->trailing == 0) || !barrier->blocking);
     put->barrier.data = NULL;
     /* If we're closing, abort the request. */
     if (uv->closing) {
@@ -635,6 +631,7 @@ int UvSnapshotPut(struct raft_io *io,
     void *cursor;
     unsigned crc;
     int rv;
+    raft_index next_index;
 
     uv = io->impl;
     assert(!uv->closing);
@@ -653,6 +650,7 @@ int UvSnapshotPut(struct raft_io *io,
     put->meta.timestamp = uv_now(uv->loop);
     put->trailing = trailing;
     put->barrier.data = put;
+    put->barrier.blocking = trailing == 0;
 
     req->cb = cb;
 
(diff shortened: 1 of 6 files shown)


MathieuBordere:
edit: there's a wonky test, I'll close for now and reopen.

src/uv_snapshot.c
```diff
@@ -677,17 +674,17 @@ int UvSnapshotPut(struct raft_io *io,
...
-            goto err_after_configuration_encode;
-        }
-    } else {
-        uvSnapshotPutStart(put);
+     * snapshot's last index + 1.
+     * - When we are only writing a snapshot during normal operation, we close
+     * all current open segments. New writes can continue on newly opened
+     * segments that will only contain entries that are newer than the snapshot,
+     * and we don't change append_next_index. */
+    raft_index next_index = trailing == 0 ? snapshot->index + 1 : uv->append_next_index;
```
freeekanayaka, line 684: Nitpick: I'd recommend to put all variable declarations at the top of the block, instead of inline. I've personally found that it helps with readability and also it reduces the risk of shadowing a variable previously declared in an outer block. That's the current convention in the code base IIRC.

freeekanayaka:
Nice catch, the approach looks good to me.

Strictly speaking perhaps requiring to close the open segment is probably necessary on for the very first snapshot that is taken (where typically there are no closed segments yet, leading to the bug you discovered), while in case there are already closed segments the bug shouldn't occur. However closing the open segment doesn't seem to cost much (compared to the overall cost of taking a snapshot), so I guess it's fine.

MathieuBordere:
> Nice catch, the approach looks good to me.
> 
> Strictly speaking perhaps requiring to close the open segment is probably necessary on for the very first snapshot that is taken (where typically there are no closed segments yet, leading to the bug you discovered), while in case there are already closed segments the bug shouldn't occur. However closing the open segment doesn't seem to cost much (compared to the overall cost of taking a snapshot), so I guess it's fine.

Thanks, an alternative was to add the start index to the open segments, but that would have been trickier I think (dealing with multiple formats, and the segment logic is already quite complex imo).

freeekanayaka:
> > Nice catch, the approach looks good to me.
> > Strictly speaking perhaps requiring to close the open segment is probably necessary on for the very first snapshot that is taken (where typically there are no closed segments yet, leading to the bug you discovered), while in case there are already closed segments the bug shouldn't occur. However closing the open segment doesn't seem to cost much (compared to the overall cost of taking a snapshot), so I guess it's fine.
> 
> Thanks, an alternative was to add the start index to the open segments, but that would have been trickier I think (dealing with multiple formats, and the segment logic is already quite complex imo).

The reason open segments don't have the start index is that you don't know that upfront (when creating the open segment file on disk) and renaming the open segment file each time you start using it slows down the throughput.

It probably goes without saying, but FWIW I do agree that the segment logic is complex, and that seems the price for its efficiency: not sure you know it, but just for reference that logic was taken virtually unchanged from [LogCabin](https://github.com/logcabin/logcabin/blob/master/Storage/SegmentedLog.h#L42), which is the very first Raft implementation, coded by the Raft author himself. IIRC etcd has a similar approach to a certain extent.

src/uv_snapshot.c
MathieuBordere: fixed

Issue #245, opened 2021-11-05
src/uv_writer.c:370: UvWriterClose: Assertion `!w->closing' failed


tomponline:
https://jenkins.linuxcontainers.org/job/lxd-github-pull-test/5139/arch=arm64,backend=dir,compiler=golang-1.16/console#console-section-5

```
/lxc-ci/build/tmp.KQmG1cbVOF/go/bin/lxc copy c1 c2 --refresh --instance-only --verbose
lxd: src/uv_writer.c:370: UvWriterClose: Assertion `!w->closing' failed.
Error: Put "http://unix.socket/1.0/instances/c2": EOF
```

MathieuBordere:
I hope https://github.com/canonical/raft/pull/265 fixes this one, if it doesn't occur in the lxd test suite for a while after this PR is merged, I will close this issue.

MathieuBordere:
This issue appeared before that PR, but the PR seems to make it occur more often (at least I was able to reproduce it, before I wasn't).

freeekanayaka:
Ah I see, good that the PR made it more evident then.

MathieuBordere:
@tomponline Have you seen this one since then?

tomponline:
Not that ive been aware of, but not been looking out for it particularly.

MathieuBordere:
Will close for now, happy to reopen if it pops up again.

PR #267, opened 2022-03-04
uv_append: Check if segment wasn't closed in write request cb

MathieuBordere:
Fixes an instance of https://github.com/canonical/raft/issues/245.

The callback of the append request being fired by the write completing, could trigger a new append that finalized the segment that was just written to, leading to the assertion being triggered. I was able to reproduce this in a couple of longer running tests, with this fix, the issue disappeared.


tomponline:
Thanks for this! :)

PR #268, opened 2022-03-23
Async snapshot

MathieuBordere:
This PR introduces:
-  a `raft_io->async_work` method that allows raft to run any workload asynchronously from the main loop.
-  raft_fsm version 2 that introduces the `snapshot_async` and `snapshot_finalize` method. `snapshot_async` is ran through the new `async_work` interface and allows a user of raft to run snapshots out of the main loop.

Updated `version-info` according to https://www.gnu.org/software/libtool/manual/html_node/Updating-version-info.html

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -150,6 +150,7 @@ libraft_la_SOURCES += \
   src/uv_tcp_listen.c \
   src/uv_tcp_connect.c \
   src/uv_truncate.c \
+  src/uv_work.c \
   src/uv_writer.c
 libraft_la_LDFLAGS += $(UV_LIBS)
 
@@ -198,7 +199,8 @@ test_integration_uv_SOURCES = \
   test/integration/test_uv_tcp_connect.c \
   test/integration/test_uv_tcp_listen.c \
   test/integration/test_uv_snapshot_put.c \
-  test/integration/test_uv_truncate.c
+  test/integration/test_uv_truncate.c \
+  test/integration/test_uv_work.c
 test_integration_uv_CFLAGS = $(AM_CFLAGS) -Wno-type-limits -Wno-conversion
 test_integration_uv_LDFLAGS = -no-install $(UV_LIBS)
 test_integration_uv_LDADD = libtest.la
--- a/Makefile.am
+++ b/Makefile.am
@@ -8,7 +8,7 @@ raftinclude_HEADERS =
 
 lib_LTLIBRARIES = libraft.la
 libraft_la_CFLAGS = $(AM_CFLAGS) -fvisibility=hidden
-libraft_la_LDFLAGS = -version-info 0:7:0
+libraft_la_LDFLAGS = -version-info 1:0:0
 libraft_la_SOURCES = \
   src/byte.c \
   src/client.c \
--- a/configure.ac
+++ b/configure.ac
@@ -1,5 +1,5 @@
 AC_PREREQ(2.60)
-AC_INIT([raft], [0.11.2])
+AC_INIT([raft], [0.12.0])
 AC_LANG([C])
 AC_CONFIG_MACRO_DIR([m4])
 AC_CONFIG_AUX_DIR([ac])
--- a/include/raft.h
+++ b/include/raft.h
@@ -418,6 +418,20 @@ struct raft_io_snapshot_get
 };
 
 /**
+ * Asynchronous work request.
+ */
+struct raft_io_async_work;
+typedef int (*raft_io_async_work_fn)(struct raft_io_async_work *req);
+typedef void (*raft_io_async_work_cb)(struct raft_io_async_work *req,
+				      int status);
+struct raft_io_async_work
+{
+    void *data;                 /* User data */
+    raft_io_async_work_fn work; /* Function to run async from the main loop */
+    raft_io_async_work_cb cb;   /* Request callback */
+};
+
+/**
  * Customizable tracer, for debugging purposes.
  */
 struct raft_tracer
@@ -499,6 +513,9 @@ struct raft_io
                         raft_io_snapshot_get_cb cb);
     raft_time (*time)(struct raft_io *io);
     int (*random)(struct raft_io *io, int min, int max);
+    int (*async_work)(struct raft_io *io,
+                      struct raft_io_async_work *req,
+                      raft_io_async_work_cb cb);
 };
 
 struct raft_fsm
--- a/include/raft.h
+++ b/include/raft.h
@@ -423,7 +423,7 @@ struct raft_io_snapshot_get
 struct raft_io_async_work;
 typedef int (*raft_io_async_work_fn)(struct raft_io_async_work *req);
 typedef void (*raft_io_async_work_cb)(struct raft_io_async_work *req,
-				      int status);
+                                      int status);
 struct raft_io_async_work
 {
     void *data;                 /* User data */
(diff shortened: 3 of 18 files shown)


freeekanayaka (commented):
I didn't go through all details yet, but here's a couple of questions:

1. If you can disclose that, where did the need of adding this feature come from? (I assume some real world use case). I ask because I'm slightly surprised that synchronous snapshots can become a performance issue, at least with the assumption that the whole FSM is kept in memory.
2. In version 2 of `raft_fsm` is it intentional that the `snapshot` method still gets called even if `snapshot_async` is defined?

MathieuBordere:
> I didn't go through all details yet, but here's a couple of questions:
> 
>     1. If you can disclose that, where did the need of adding this feature come from? (I assume some real world use case). I ask because I'm slightly surprised that synchronous snapshots can become a performance issue, at least with the assumption that the whole FSM is kept in memory.
> 
>     2. In version 2 of `raft_fsm` is it intentional that the `snapshot` method still gets called even if `snapshot_async` is defined?

1. Yes, with larger databases they can become an issue, on my system a database of 1.5GB can take 800-900ms to snapshot, which is quite long to block the eventloop.

We have a (rudimentary) benchmark, a typical result of a run with synchronous snapshots on the Github actions machines is something like 

```
Write results:
n 677104
n_err 1
avg [ms] 4.759796
max [ms] 17417.409460
min [ms] 0.354293

Read results:
n 676192
n_err 0
avg [ms] 0.438674
max [ms] 12199.753818
min [ms] 0.038299
```

Notice, the large maximum times a write or a read can take.

A typical run with async snapshots is (on the dqlite branch where I'm implementing it)

```
Write results:
n 716654
n_err 1
avg [ms] 4.886415
max [ms] 4860.268703
min [ms] 0.487495

Read results:
n 715646
n_err 0
avg [ms] 0.108416
max [ms] 536.435864
min [ms] 0.044499
```

The spikes are still there, but a lot less high. I'll try to do some more measurements tomorrow.

2. Yes, it's the intention `snapshot` is called in the main loop, in dqlite e.g. this is used to copy the wal and prevent checkpoints from running, Afterwards `snapshot_async` is run in a worker thread, in dqlite we would copy pointers to the database pages to raft buffers in this method. And finally, `snapshot_finalize` is called in the main loop again, in dqlite that will allow checkpoints to run again and free any allocated memory.

MathieuBordere:
> Oh I see. So you'd copy the whole WAL synchronously (since it should typically be relatively small) so there are no race conditions, and then copy the database part asynchronously and since it won't be changed (checkpoints off) it is okay to do that in a separate thread.

Yes, and now that checkpoints are off and we have a copy of the WAL, it's no longer needed to copy the database, we can just pass the page pointers to raft.

freeekanayaka:
> > Oh I see. So you'd copy the whole WAL synchronously (since it should typically be relatively small) so there are no race conditions, and then copy the database part asynchronously and since it won't be changed (checkpoints off) it is okay to do that in a separate thread.
> 
> Yes, and now that checkpoints are off and we have a copy of the WAL, we no longer need to copy the database, we can just pass the page pointers to raft.

Ah nice.

freeekanayaka:
Another question: since only page pointers are passed to raft, what would the `snapshot_async` implementation of dqlite do that is expensive and requires to be run outside of the main loop? It feels that basically the real trick here is to disable checkpoints, so in principle just `snapshot_finalize` would be needed? Or I guess I'm missing something.

MathieuBordere:
> Another question: since only page pointers are passed to raft, what would the `snapshot_async` implementation of dqlite do that is expensive and requires to be run outside of the main loop? It feels that basically the real trick here is to disable checkpoints, so in principle just `snapshot_finalize` would be needed? Or I guess I'm missing something.

Yeah indeed, I first made this while I still did the copy of the database in the worker thread, only afterwards I moved dqlite to passing pointers. I still think it's a bit cleaner to offload the work to a thread.

MathieuBordere:
> In which sense do you think it's cleaner? I'd say it's cleaner if the goal is "add support for asynchronous snapshots". But if we switch the goal to "add support for post-snapshot cleanup", then simply introducing `snapshot_finalize` would be equally clean.
> 
> I'm just a bit perplexed because it feels there's additional complexity (new `raft_io` method, new `raft_io` request, new snapshot-related code paths etc) that is not strictly needed right now to solve the performance issue of dqlite.
> 
> I know it might sound bad, but perhaps it'd be worth considering implementing just `snapshot_finalize` instead for now (i.e. support for snapshot cleanup)? And keep this branch around as experimental until there's need for it in dqlite or someone asks for it for some other use case.

It was already merged in this night, if you really object, I'll talk to @stgraber to see if it's worth to revert it.
I would need to do some measurements to find out what happens if we would do all the work in the main thread instead of in a worker.

freeekanayaka:
Assuming that the work is copying pointers it should really be fast and I wouldn't expect any regression.

I'd personally go for the minimum required change (adding `snapshot_finalize`), the main reason being that adding parallelism can always introduce subtle issues that might not be immediate to see right now, and it's also something to maintain and be careful of when modifying the code in the future. But I'll let you guys decide since YMMV.

freeekanayaka:
For example, we have a few spots that check if a snapshot is in progress, and they use `if (r->snapshot.put.data != NULL)`, e.g.: 

https://github.com/canonical/raft/blob/master/src/replication.c#L1276
https://github.com/canonical/raft/blob/master/src/replication.c#L1418
https://github.com/canonical/raft/blob/master/src/replication.c#L1657

Do they still work correctly now that a snapshot might be in progress even if `r->snapshot.put.data` is `NULL` (because we're still running `snapshot_async`)? Or am I misinterpreting the change?

MathieuBordere:
> For example, we have a few spots that check if a snapshot is in progress, and they use `if (r->snapshot.put.data != NULL)`, e.g.:
> 
> https://github.com/canonical/raft/blob/master/src/replication.c#L1276 https://github.com/canonical/raft/blob/master/src/replication.c#L1418 https://github.com/canonical/raft/blob/master/src/replication.c#L1657
> 
> Do they still work correctly now that a snapshot might be in progress even if `r->snapshot.put.data` is `NULL` (because we're still running `snapshot_async`)? Or am I misinterpreting the change?

https://github.com/canonical/raft/blob/master/src/replication.c#L1276 check if `snapshot.pending.term != 0` to detect if a snapshot is in progress, that's still okay.

same for https://github.com/canonical/raft/blob/master/src/replication.c#L1418

https://github.com/canonical/raft/blob/master/src/replication.c#L1657 only tries to detect a snapshot install and is still okay.

MathieuBordere:
Don't understand me wrong, I understand your concern, and I would probably have done it differently after your input.

MathieuBordere:
Did some measurements and the async work is indeed really fast, I would probably then revert this merge and work on the simpler approach to remove some risk.

freeekanayaka:
Thanks, sounds like a wise choice to me.

MathieuBordere:
> Thanks, sounds like a wise choice to me.

Thanks for the input.

calvin2021y:
hi @MathieuBordere 

Can we have a async branch with updated to lasted fixed commit ?

I am work on a SQL based solution with in memory database, I need async snapshot so I can flush the database into disk.

Without async snapshot it will block the event look for long time If I have a big database.

MathieuBordere:
> hi @MathieuBordere
> 
> Can we have a async branch with updated commit to master ?
> 
> I am work on a SQL based solution with in memory database, I need async snapshot so I can flush the database into disk.
> 
> Without async snapshot it will block the event loop for long time If I have a big database.
> 
> I also need async restore with a callback let the raft know it is ready to continue. (without block the loop, because there could be raft group run in it)

Hi, it will probably land this week but you can already find it [here](https://github.com/MathieuBordere/raft/tree/async-ss) .

calvin2021y:
hi @MathieuBordere

I has merge your branch into master and it work well.  

please consider add async_restore.

With async_restore I am able to add disk support with fast snapshot for big size database.   Without async_restore the event will blocked and break a lot others.  (like NATS sub client run in the event loop, or other raft group)

PR #269, opened 2022-03-24
Revert "Async snapshot"

MathieuBordere:
After discussion with @freeekanayaka I'll opt for a simpler approach and just introduce `snapshot_finalize`.

Reverts canonical/raft#268


MathieuBordere:
The cla-check fails because I used Github's functionality to create this revert, I don't know if it bothers you @stgraber , I can recreate it properly if you want to.

PR #270, opened 2022-03-29
Snapshot finalize

MathieuBordere:
- Introduce `raft_fsm` version 2 adding `snapshot_finalize`. `snapshot_finalize` allows an implementer to clean up state set by `snapshot`. When `snapshot_finalize` is set, raft assumes no ownership of any of the `raft_buffer`s in the snapshot, the implementer must make sure they are cleaned up. 
In `dqlite` this will be used like this:
`snapshot`: copy WAL, disallow checkpoints from running, copy pointers to database pages to `raft_buffer`s (shallow copy, only pointers are copied, less memory allocation + faster). Not allowing checkpoints allows raft to safely access the database pages without the risk of them being changed. `dqlite` writers can still append to the WAL, as we have made a copy.
`snapshot_finalize`: allow checkpoints to run again, free allocated buffers.

- Bumped version twice because the previously reverted async snapshot PR already bumped the version once.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -8,7 +8,7 @@ raftinclude_HEADERS =
 
 lib_LTLIBRARIES = libraft.la
 libraft_la_CFLAGS = $(AM_CFLAGS) -fvisibility=hidden
-libraft_la_LDFLAGS = -version-info 0:7:0
+libraft_la_LDFLAGS = -version-info 2:0:0
 libraft_la_SOURCES = \
   src/byte.c \
   src/client.c \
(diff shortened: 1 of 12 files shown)


freeekanayaka (approved):

freeekanayaka:
Looks good to me, thanks, it feels like a simpler approach.

MathieuBordere:
> Looks good to me, thanks, it feels like a simpler approach.

Yeah, it's a lot simpler, I like it, thank you.

PR #274, opened 2022-04-07
Add raft_uv_tcp_set_bind_address().

### LINKED ISSUE #273: Socket binding issue inside containers
Hello,

Thanks for a very handy library.

I'm running into a problem when trying to run my application that uses raft inside a container. Calling `raft_init(..., "192.168.1.1:8080")` doesn't work because the container IP address is not the same as the external IP address. I tried using `raft_init(..., "0.0.0.0:8080")` and that worked when running multiple local instances outside of a container, but not inside the container.

I got a working result with this patch:

```
diff --git a/src/uv_tcp_listen.c b/src/uv_tcp_listen.c
index d11f84b..1386af4 100644
--- a/src/uv_tcp_listen.c
+++ b/src/uv_tcp_listen.c
@@ -284,7 +284,7 @@ int UvTcpListen(struct raft_uv_transport *transport, raft_uv_accept_cb cb)
     t = transport->impl;
     t->accept_cb = cb;
 
-    rv = uvIpParse(address, &addr);
+    rv = uvIpParse("0.0.0.0:8080", &addr);
     if (rv != 0) {
         return rv;
     }
```

This binds the listener to 0.0.0.0:8080, which means the port can happily be mapped to, and the server address of 192.168.1.1 is still available internally for the rest of the library.

I recognise this isn't a useful solution at the moment. I did consider making a modified version of `uvIpParse()` that would ignore the host part, but wanted your thoughts on this. I can see that some people would want to to be able to bind to a specific interface address so changing to using 0.0.0.0 across the board isn't great.

Perhaps there could be an option to set a bind address to be called before `raft_init()`. When `raft_init()` was called, it would use the bind address if set, otherwise use the address provided to `raft_init()`?

ralight:
This allows an application to choose which network interface the
listening socket will bind to.

Closes #273.

Signed-off-by: Roger A. Light <roger@atchoo.org>

### DIFF
--- a/src/uv_tcp.c
+++ b/src/uv_tcp.c
@@ -1,3 +1,4 @@
+#include "uv_ip.h"
 #include "uv_tcp.h"
 
 #include <string.h>
@@ -6,6 +7,7 @@
 #include "../include/raft/uv.h"
 #include "assert.h"
 #include "err.h"
+#include "heap.h"
 
 /* Implementation of raft_uv_transport->init. */
 static int uvTcpInit(struct raft_uv_transport *transport,
@@ -75,6 +77,7 @@ int raft_uv_tcp_init(struct raft_uv_transport *transport,
     t->loop = loop;
     t->id = 0;
     t->address = NULL;
+    t->bind_address = NULL;
     t->listener.data = NULL;
     t->accept_cb = NULL;
     QUEUE_INIT(&t->accepting);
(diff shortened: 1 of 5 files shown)


src/uv_tcp.c
```diff
@@ -97,3 +99,22 @@ void raft_uv_tcp_close(struct raft_uv_transport *transport)
...
+    struct UvTcp *t = transport->impl;
+    struct sockaddr_in addr;
+    int rv;
+
+    rv = uvIpParse(address, &addr);
+    if (rv != 0) {
+        return RAFT_INVALID;
+    }
+
+    t->bind_address = strdup(address);
```
MathieuBordere, line 115: This memory is never freed, we also usually use the allocator functions from `src/heap.c`, you could do something similar as here https://github.com/canonical/raft/blob/3769e75e5adbb6954c557403818cb53cc9aa14e1/src/raft.c#L44

MathieuBordere (changes requested):
Hi, thanks for this!

Could you fix the memory leak and add a basic test for this in https://github.com/canonical/raft/blob/master/test/integration/test_uv_tcp_listen.c ?

ralight:
Thanks. I've fixed the memory, but could do with a hint on what you'd like out of the test? I could repeat the `/* If the handshake is successful, the accept callback is invoked. */` test, but with a custom `setUp()` function?

MathieuBordere:
> Thanks. I've fixed the memory, but could do with a hint on what you'd like out of the test? I could repeat the `/* If the handshake is successful, the accept callback is invoked. */` test, but with a custom `setUp()` function?

Yes, that would be fine. I see that the call to `raft_init` is buried pretty deep in the test setup and you would probably need to touch a lot of code to make it work, I'm not sure if it's worth it. @freeekanayaka What do you think?

ralight:
I've added a test which covers the call. It's not especially satisfactory, let me know what you think.

PR #277, opened 2022-06-02
Startup recovery

MathieuBordere:
This PR attempts to move corrupt segment files out of the way at startup. Once moved out of the way, the next startup can complete successfully requiring no manual intervention. The renamed files are named "corrupt-{timestamp}-{filename}" and are kept around in case they would contain useful data.

The logic is as follows:
- open segments: Are basically only renamed in case of an invalid format header, the other cases are already covered. If an open segment is corrupt, the other open segments are renamed too. In case they contained valid data, raft wouldn't attempt to load the data, because loading it could lead to missing transactions that were part of the corrupt segment leading to an inconsistent state of the log.
- closed segments: Are renamed when they are corrupt and if they are the last closed segment file that is found. When they are renamed all open segments are renamed too, for the same reason as above.

This would lead to less manual intervention in projects that make use of libraft, the trade-off is less visibility of the corruption.

### DIFF
--- a/src/uv_segment.c
+++ b/src/uv_segment.c
@@ -1,4 +1,6 @@
 #include <errno.h>
+#include <inttypes.h>
+#include <stdint.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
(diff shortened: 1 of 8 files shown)


stgraber:
@freeekanayaka can you take a quick look at this one?

freeekanayaka:
> @freeekanayaka can you take a quick look at this one?

Yes, definitely, will do over the weekend.

test/integration/test_uv_load.c
```diff
@@ -883,6 +924,196 @@ TEST(load, openSegmentNoClosedSegmentsSnapshotPresent, setUp, tearDown, 0, NULL)
...
+    LOAD_NO_SETUP(0,         /* term */
+                  0,         /* voted for */
+                  &snapshot, /* snapshot */
+                  4,         /* start index */
+                  4,         /* data for first loaded entry */
+                  2          /* n entries */
+    );
+    return MUNIT_OK;
+}
+/* The data directory contains a closed segment and an open segment with a corrupt
```
freeekanayaka, line 1028: Nit: missing blank line.

freeekanayaka:
Looks good to me, since it automates what we have already been telling people to do manually (AFAIUK).

Maybe it would be nice to:

1. Not enable this behavior by default, and have it opt-in instead (since there might be consumers of libraft that don't want it). Or alternatively have it enabled by default but offer the possibility to opt-out. For example we could add a new ` void raft_uv_set_auto_recovery(struct raft_io *io, bool flag)` public function to `raft/uv.h` for that.
2. In case the auto-recovery behavior is enabled, maybe there's no point to still return a `RAFT_CORRUPT` error? The logic could be modified to make the recovery completely transparent and not require a second startup.

These points are optional, YMMV.

freeekanayaka:
Also not that without point 2. in place, I'm not sure what a user is supposed to do when it gets back a `RAFT_CORRUPT` error, as it has now way to know if "retrying" could help or not. It could retry just once and give up immediately and if that retry fails, but it feels a bit weird. Or maybe I'm missing the bigger picture of how this is supposed to be consumed.

MathieuBordere:
> Also not that without point 2. in place, I'm not sure what a user is supposed to do when it gets back a `RAFT_CORRUPT` error, as it has now way to know if "retrying" could help or not. It could retry just once and give up immediately and if that retry fails, but it feels a bit weird. Or maybe I'm missing the bigger picture of how this is supposed to be consumed.

I think it makes sense to try and recover without the restart, thanks.

MathieuBordere:
PR has been adapted to automatically try to recover once if a corrupted segment was encountered. This behavior is enabled by default but can be turned off.

I've duplicated most of the tests from the original PR to run them with and without recovery, might look a bit verbose.

src/uv_segment.c
```diff
@@ -801,6 +803,69 @@ void uvSegmentBufferReset(struct uvSegmentBuffer *b, unsigned retain)
...
+        tracef("%s", errmsg);
+        return;
+    }
+}
+
+/*
+ * On startup, raft will try to recover when a corrupt segment is detected.
+ *
+ * When a corrupt open segment is encountered, it, and all subsequent open segments,
+ * , are renamed. Not renaming newer, possible non-corrupt, open segments could lead to loading
```
freeekanayaka, line 839: There's an extra "`, `" at the beginning of the line.

freeekanayaka (approved):
Looks good to me, thanks.

PR #280, opened 2022-06-09
Prevote term

MathieuBordere:
Fixes a bug where a cluster with pre-vote enabled remained unavailable when there was a quorum of nodes online but where one of the nodes has an outdated term resulting in the other servers ignoring its vote.
This was caused by the fact that a node did not increase its term when receiving a pre-vote request, whereas the node should have checked if it could update its term to `pre_vote request.term - 1`, the actual term of the candidate server.

The newly added test `election/changeReviveOutdated` would fail with cluster sizes `3` and `4` and `pre-vote=1` without the fix in the last commit.

Also fixes a bug in the fuzzy election test that resulted in never running the tests with pre-vote disabled.

log of failing test
```
Running test suite with seed 0xfe5526c2...
election/changeReviveOutdated                               
  cluster-n=4, cluster-pre-vote=1                           [ ERROR ]
1:                    src/start.c:143 - starting

REMOVED BOILER PLATE LOGS

4:                  src/convert.c: 23 - old_state:0 new_state:1

SERVER 4 IS KILLED

2:                     src/tick.c: 44 - convert to candidate and start new election
2:                  src/convert.c: 38 - clear follower state
2:                  src/convert.c: 23 - old_state:1 new_state:2
3:        src/recv_request_vote.c: 32 - self:3 from:2@2 candidate_id:2 disrupt_leader:0 last_log_index:1 last_log_term:1 pre_vote:1 term:2
3:                 src/election.c:259 - remote log equal or longer than local -> granting vote
3:                 src/election.c:280 - vote granted to 2
1:        src/recv_request_vote.c: 32 - self:1 from:2@2 candidate_id:2 disrupt_leader:0 last_log_index:1 last_log_term:1 pre_vote:1 term:2
1:                 src/election.c:259 - remote log equal or longer than local -> granting vote
1:                 src/election.c:280 - vote granted to 2
2: src/recv_request_vote_result.c: 27 - self:2 from:3@3 term:1 vote_granted:1 pre_vote:1
2: src/recv_request_vote_result.c:136 - votes quorum not reached
2: src/recv_request_vote_result.c: 27 - self:2 from:1@1 term:1 vote_granted:1 pre_vote:1
2: src/recv_request_vote_result.c:120 - votes quorum reached -> pre-vote successful
3:        src/recv_request_vote.c: 32 - self:3 from:2@2 candidate_id:2 disrupt_leader:0 last_log_index:1 last_log_term:1 pre_vote:0 term:2
[...]

Issue #279, opened 2022-06-08
Possible pre-vote bug related bug


ralight:
I think I've found a bug when using pre-voting and would like your thoughts.

I've created a PR which has a modified example that makes it easier to see: https://github.com/canonical/raft/pull/278

Manually start four nodes in four terminals:

```
mkdir 1 2 3 4
./server 1 1
./server 2 2
./server 3 3
./server 4 4
```

One of the nodes will be elected leader, I am assuming it is node 1. Kill one of the followers, e.g. 2, then kill the leader, e.g. 1. The remaining two nodes, 3 and 4, will have an election and rightly not be able to proceed as they do not have quorum.

Now restart node 2. This should allow a leader to be elected. The node 2 will have a term which is behind that of the nodes 3 and 4. It does not update its term in response to communicating with the nodes 3 and 4. This means that nodes 3 and 4 will exchange votes, but the vote from node 2 will be rejected as it has an earlier term. The result is that quorum is not reached and so a new election is called.

Now restart node 1, and everything will work again.

 If I disable pre-vote and repeat the above procedure, then the three node version of the cluster manages to achieve quorum despite node 2 restarting with an earlier term.

Have I misunderstood what should be happening with pre-vote, or is this a bug?

I can provide logs if you would like, I didn't want to clutter up the report in the first instance though.

MathieuBordere:
Logs would be handy, as it's not clear to me what exactly the bug would be.

`pre-vote` is there to try and prevent a functioning cluster to be disrupted when a node that was disconnected from the cluster is reconnected to the cluster with an out-of-date log, increases its term but loses the election because its log is not up to date. As a result new election(s) is/are started that will be won by a node with an up-to-date log, but in meanwhile the cluster was unavailable. `pre-vote` is there to probe if a node has a chance to win an election, only if it has a chance will it increase it's term and start a real election.

ralight:
Node 1:

```
$ !./
./server 1 1
1: starting
starting (start.c:143)
current_term:1 voted_for:0 start_index:1 n_entries:1 (start.c:151)
restore 1 entries starting at 1 (start.c:181)
restore most recent configuration (configuration.c:342)
=== CONFIG START === (configuration.c:343)
id:1 address:127.0.0.1:9001 role:1 (configuration.c:348)
id:2 address:127.0.0.1:9002 role:1 (configuration.c:348)
id:3 address:127.0.0.1:9003 role:1 (configuration.c:348)
[...]

PR #281, opened 2022-06-10
Prevote mirror term

MathieuBordere:
Based on the solution used in ScyllaDB [1] to prevent a cluster deadlock
when a node with a stale term reconnects to the cluster but its vote
is needed to form quorum. Using the stale term in the response would
lead to the other nodes ignoring the response and would result in an
unavailable cluster.

[1]: https://github.com/scylladb/scylla/blob/e5956fee8af8ad7b6ba1e797f269729406afd7ef/raft/fsm.cc#L794

Reverts the previous solution to the same problem stated in https://github.com/canonical/raft/issues/279 and fixed in https://github.com/canonical/raft/pull/280

### DIFF
--- a/src/recv_request_vote.c
+++ b/src/recv_request_vote.c
@@ -21,7 +21,6 @@ int recvRequestVote(struct raft *r,
     struct raft_io_send *req;
     struct raft_message message;
     struct raft_request_vote_result *result = &message.request_vote_result;
-    raft_term check_term;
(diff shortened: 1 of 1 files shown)


freeekanayaka (approved):
The change feels simpler and more elegant, thanks. Nice docstring and good to see that the tests that were put in place in the previous PR still pass.

PR #284, opened 2022-06-22
Deprecate raft_fixture_init

MathieuBordere:
Accessing the client-supplied `struct raft_fsm` array can cause memory
errors when the client was compiled against a raft version with an
older, smaller `struct raft_fsm` version when using `raft_fixture_init`.
This happens because the new raft library will stride through the array with the new size,
while the binary, compiled against an older version, will only allocate space
for the old size, this results in corruptions.

### DIFF
--- a/include/raft.h
+++ b/include/raft.h
@@ -502,20 +502,6 @@ struct raft_io
 };
 
 /*
- * version 1:
- * struct raft_fsm
- * {
- *     int version;
- *     void *data;
- *     int (*apply)(struct raft_fsm *fsm,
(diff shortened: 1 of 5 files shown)


stgraber:
I'm not particularly fond of public symbols having `_v2` or the like in their name, can we have a more descriptive name instead?

calvin2021y:
maybe add raft_fixture_init_fixed_size, and pass the version or raft_fsm size into it.

freeekanayaka:
LGTM, +1

PR #286, opened 2022-06-28
Async snapshots

MathieuBordere:
Same functionality as in (reverted) https://github.com/canonical/raft/pull/268 , but decided to open a new PR for clarity.

As discussed, this is needed for dqlite's disk-mode, copying a multi GB database in the main loop would be disastrous.
I will not be porting dqlite's regular memory operation mode to async snapshots as it's not needed imo and limits some risks.

### DIFF
--- a/src/raft.c
+++ b/src/raft.c
@@ -24,6 +24,10 @@
 #define DEFAULT_MAX_CATCH_UP_ROUNDS 10
 #define DEFAULT_MAX_CATCH_UP_ROUND_DURATION (5 * 1000)
 
+static int ioFsmCompat(struct raft *r,
+                       struct raft_io *io,
+                       struct raft_fsm *fsm);
+
 int raft_init(struct raft *r,
               struct raft_io *io,
               struct raft_fsm *fsm,
@@ -32,6 +36,12 @@ int raft_init(struct raft *r,
 {
     int rv;
     assert(r != NULL);
+
+    rv = ioFsmCompat(r, io, fsm);
+    if (rv != 0) {
+	goto err;
+    }
+
     r->io = io;
     r->io->data = r;
     r->fsm = fsm;
@@ -227,3 +237,18 @@ unsigned long long raft_digest(const char *text, unsigned long long n)
 
     return byteFlip64(digest);
 }
+
+static int ioFsmCompat(struct raft *r,
+                       struct raft_io *io,
+                       struct raft_fsm *fsm)
+{
+    if ((fsm->version > 2 && fsm->snapshot_async != NULL)
+        && ((io->version < 2) || (io->async_work == NULL))) {
+        ErrMsgPrintf(r->errmsg,
+            "async snapshot requires io->version > 1 and async_work method.");
+        return -1;
+    }
+
+    return 0;
+}
+
--- a/src/replication.c
+++ b/src/replication.c
@@ -1455,10 +1455,56 @@ static void takeSnapshotCb(struct raft_io_snapshot_put *req, int status)
     }
 
     logSnapshot(&r->log, snapshot->index, r->snapshot.trailing);
-
 out:
-    takeSnapshotClose(r, &r->snapshot.pending);
+    takeSnapshotClose(r, snapshot);
     r->snapshot.pending.term = 0;
+}
+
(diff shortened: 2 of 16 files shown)


src/raft.c
```diff
@@ -24,6 +24,10 @@
 #define DEFAULT_MAX_CATCH_UP_ROUNDS 10
 #define DEFAULT_MAX_CATCH_UP_ROUND_DURATION (5 * 1000)
 
+static int raft_io_fsm_compat(struct raft *r,
```
freeekanayaka, line 27: Nit: should be camel-case (e.g. `ioFsmCompat` or `raftIoFsmCompat`), since it's a private function.

src/replication.c
```diff
@@ -1489,17 +1536,30 @@ static int takeSnapshot(struct raft *r)
...
 
-    assert(r->snapshot.put.data == NULL);
-    r->snapshot.put.data = r;
-    rv = r->io->snapshot_put(r->io, r->snapshot.trailing, &r->snapshot.put,
-                             snapshot, takeSnapshotCb);
-    if (rv != 0) {
-        goto abort_after_fsm_snapshot;
+    bool sync_snapshot = r->fsm->version < 3 || r->fsm->snapshot_async == NULL;
+    if (sync_snapshot) {
+        /* putSnapshot will clean up config and buffers case of error */
```
freeekanayaka, line 1541: s/buffers case/buffers in case/

src/replication.c
```diff
@@ -1489,17 +1536,30 @@ static int takeSnapshot(struct raft *r)
...
+        return putSnapshot(r, snapshot, takeSnapshotCb);
+    } else {
+        struct raft_io_async_work *take = raft_malloc(sizeof(*take));
+        if (take == NULL) {
+            rv = RAFT_NOMEM;
+            goto abort_after_snapshot;
+        }
+        take->cb = NULL;
```
freeekanayaka, line 1549: Nit: I'd avoid setting the `cb` field since it's essentially a "private" field reserved for the `raft_io` implementation. The "user" fields are really just `data` and `work`. Not setting it also makes code a tad more clear since you're passing a `takeSnapshotDoneCb` below, and the two callbacks should not be confused.

freeekanayaka (approved):

freeekanayaka:
I'm sure that you've already considered that, but just for clarity: the I/O for persisting the snapshot will be async, but still the snapshot itself needs to be loaded completely in memory. Not sure how much of a problem will it be for the dqlite use cases that require that the database does not live in memory (presumably taking snapshots will result in memory spikes).

calvin2021y:
Thanks for the great work.

As @freeekanayaka pointer out, async snapshot will not work well without async restore. 

If leader ask node install snapshot then event loop blocked. ( for example a offline node come back online)

Or run more than one raft in event loop, then event loop will blocked during restore process. the other raft instance will timeout.

With async restore, you can replace the state machine instance use atomic CAS,  or just replace it without lock from main event loop.

freeekanayaka:
I think those are all fine approaches, ideally there should be some plan about it so we avoid to introduce API changes here that would need further adaptation. It's not a deal breaker, since we're not yet at `1.0` and we're not making any promise about stability.

MathieuBordere:
I'll devise a plan before merging this.

freeekanayaka:
One thought I've always had since the goal of storing dqlite's db on disk came out, is that perhaps it might make sense to basically use directly raft snapshots as on-disk data source. They already are essentially SQLite database images, and if dqlite starts using the disk also the checkpoint operation becomes virtually equivalent to taking snapshot and trim the raft log.

It's a thought that is bit orthogonal to this PR, but perhaps it's worth considering it.

MathieuBordere:
> One thought I've always had since the goal of storing dqlite's db on disk came out, is that perhaps it might make sense to basically use directly raft snapshots as on-disk data source. They already are essentially SQLite database images, and if dqlite starts using the disk also the checkpoint operation becomes virtually equivalent to taking snapshot and trim the raft log.
> 
> It's a thought that is bit orthogonal to this PR, but perhaps it's worth considering it.

Also completely orthogonal to this PR, is that I also thought that maybe raft could be more tightly integrated with dqlite, i.e. not as a shared library, but basically part of dqlite.

MathieuBordere:
this can be merged

PR #301, opened 2022-08-29
Add additional async getaddrinfo step to outgoing tcp connections

NorbertHeusser:
As first step to allow usage of hostnames instead of static IPs I added the required additional async getaddrinfo step to the outgoing tcp connection to the other cluster members.
Unfortunately the libuv async getaddrinfo is not connected to the TCP handle. As a consequence abort the connect got a little bit more complicated as the getaddrinfo may be in progress and cannot be stopped. In this case the data structures need to be valid until the getaddrinfo callback is finished.

In the current first implementation the getaddrinfo is restricted to IPv4 addresses in a string representation as before. This way it should behave excactly as before.

### DIFF
--- a/src/uv_ip.c
+++ b/src/uv_ip.c
@@ -7,6 +7,52 @@
 
 #include "uv_ip.h"
 
+static const char *strCpyUntil(char *target,
+                               const char *source,
+                               size_t target_size,
+                               char separator)
+{
+    size_t i;
+    for (i = 0; i < target_size; ++i) {
+        if (!source[i] || source[i] == separator) {
+            target[i] = 0;
+            return source + i;
+        } else {
+            target[i] = source[i];
+        }
+    }
+    return NULL;
+}
+
+int uvIpAddrSplit(const char *address,
+                  char *host,
+                  size_t host_size,
+                  char *service,
+                  size_t service_size)
+{
+    char colon = ':';
+    const char *service_ptr = NULL;
+
+    if (host) {
+        service_ptr = strCpyUntil(host, address, host_size, colon);
+        if (!service_ptr) {
+            return RAFT_NAMETOOLONG;
+        }
+    }
+    if (service) {
+        if (!service_ptr) {
+            service_ptr = strchr(address, colon);
+        }
+        if (!service_ptr || *service_ptr == 0 || *(++service_ptr) == 0) {
+            service_ptr = "8080";
+        }
+        if (!strCpyUntil(service, service_ptr, service_size, 0)) {
+            return RAFT_NAMETOOLONG;
+        }
+    }
+    return 0;
+}
+
 int uvIpParse(const char *address, struct sockaddr_in *addr)
 {
     char buf[256];
@@ -17,7 +63,7 @@ int uvIpParse(const char *address, struct sockaddr_in *addr)
     int rv;
 
     /* TODO: turn this poor man parsing into proper one */
-    n = sizeof(buf)-1;
+    n = sizeof(buf) - 1;
     strncpy(buf, address, n);
     buf[n] = '\0';
     host = strtok(buf, colon);
--- a/src/uv_tcp_connect.c
+++ b/src/uv_tcp_connect.c
@@ -25,14 +25,16 @@
 /* Hold state for a single connection request. */
 struct uvTcpConnect
 {
-    struct UvTcp *t;             /* Transport implementation */
-    struct raft_uv_connect *req; /* User request */
-    uv_buf_t handshake;          /* Handshake data */
-    struct uv_tcp_s *tcp;        /* TCP connection socket handle */
-    struct uv_connect_s connect; /* TCP connection request */
-    struct uv_write_s write;     /* TCP handshake request */
-    int status;                  /* Returned to the request callback */
-    queue queue;                 /* Pending connect queue */
+    struct UvTcp *t;                     /* Transport implementation */
+    struct raft_uv_connect *req;         /* User request */
+    uv_buf_t handshake;                  /* Handshake data */
+    struct uv_tcp_s *tcp;                /* TCP connection socket handle */
+    struct uv_getaddrinfo_s getaddrinfo; /* DNS resolve request */
+    struct uv_connect_s connect;         /* TCP connection request */
+    struct uv_write_s write;             /* TCP handshake request */
+    uv_check_t delayedtcpclose;          /* A check handle required to delay closing tcp */
+    int status;                          /* Returned to the request callback */
+    queue queue;                         /* Pending connect queue */
 };
 
 /* Encode an handshake message into the given buffer. */
@@ -64,7 +66,9 @@ static void uvTcpConnectFinish(struct uvTcpConnect *connect)
     struct raft_uv_connect *req = connect->req;
     int status = connect->status;
     QUEUE_REMOVE(&connect->queue);
+    uv_close((struct uv_handle_s *)&connect->delayedtcpclose, NULL);
     RaftHeapFree(connect->handshake.base);
+    uv_freeaddrinfo(connect->getaddrinfo.addrinfo);
     raft_free(connect);
     req->cb(req, stream, status);
 }
@@ -83,12 +87,29 @@ static void uvTcpConnectUvCloseCb(struct uv_handle_s *handle)
     UvTcpMaybeFireCloseCb(t);
 }
 
+static void uvTcpConnectCheckMayClose(uv_check_t *handle)
+{
+    struct uvTcpConnect *connect = handle->data;
+    if (connect->status || uv_is_active((uv_handle_t *)connect->tcp)) {
+        uv_check_stop(&connect->delayedtcpclose);
+        uv_close((struct uv_handle_s *)connect->tcp, uvTcpConnectUvCloseCb);
+    }
+}
+
 /* Abort a connection request. */
 static void uvTcpConnectAbort(struct uvTcpConnect *connect)
 {
     QUEUE_REMOVE(&connect->queue);
     QUEUE_PUSH(&connect->t->aborting, &connect->queue);
-    uv_close((struct uv_handle_s *)connect->tcp, uvTcpConnectUvCloseCb);
+    if (uv_cancel((struct uv_req_s *)&connect->getaddrinfo) == 0 ||
+        !uv_is_active((uv_handle_t *)connect->tcp)) {
+        /* If canceling the addrinfo call was not successfull, but the tcp
+           handle is not active the getaddrinfo is in progress and we need to
+           delay closing the tcp handle until it's finished */
+        uv_check_start(&connect->delayedtcpclose, uvTcpConnectCheckMayClose);
+    } else {
+        uv_close((struct uv_handle_s *)connect->tcp, uvTcpConnectUvCloseCb);
+    }
 }
 
 /* The handshake TCP write completes. Fire the connect callback. */
@@ -146,23 +167,62 @@ err:
     uvTcpConnectAbort(connect);
 }
 
+static void uvGetAddrInfoCb(uv_getaddrinfo_t *req,
+                            int status,
+                            struct addrinfo *res)
+{
+    struct uvTcpConnect *connect = req->data;
+    struct UvTcp *t = connect->t;
+    int rv;
+
+    if (t->closing || status == UV_ECANCELED) {
+        connect->status = RAFT_CANCELED;
+        return;
+    }
+
+    if (status < 0) {
+        ErrMsgPrintf(t->transport->errmsg, "uv_getaddrinfo(): %s",
+                     uv_err_name(status));
+        connect->status = RAFT_NOCONNECTION;
+        goto err;
+    }
+    rv = uv_tcp_connect(&connect->connect, connect->tcp,
+                        (const struct sockaddr *)res->ai_addr,
+                        uvTcpConnectUvConnectCb);
+    if (rv != 0) {
+        /* UNTESTED: since parsing succeed, this should fail only because of
+         * lack of system resources */
+        ErrMsgPrintf(t->transport->errmsg, "uv_tcp_connect(): %s",
+                     uv_strerror(rv));
+        connect->status = RAFT_NOCONNECTION;
+        goto err;
+    }
+
+    return;
+
+err:
+    uvTcpConnectAbort(connect);
+}
 /* Create a new TCP handle and submit a connection request to the event loop. */
 static int uvTcpConnectStart(struct uvTcpConnect *r, const char *address)
 {
+    static struct addrinfo hints = {
+        .ai_flags = AI_V4MAPPED | AI_ADDRCONFIG | AI_NUMERICHOST,
+        .ai_family = AF_INET,
+        .ai_socktype = SOCK_STREAM,
+        .ai_protocol = 0};
     struct UvTcp *t = r->t;
-    struct sockaddr_in addr;
+    char hostname[NI_MAXHOST];
+    char service[NI_MAXSERV];
     int rv;
 
-    rv = uvIpParse(address, &addr);
-    if (rv != 0) {
-        goto err;
-    }
+    r->handshake.base = NULL;
 
     /* Initialize the handshake buffer. */
     rv = uvTcpEncodeHandshake(t->id, t->address, &r->handshake);
     if (rv != 0) {
         assert(rv == RAFT_NOMEM);
-        ErrMsgOom(r->t->transport->errmsg);
+        ErrMsgOom(t->transport->errmsg);
         goto err;
     }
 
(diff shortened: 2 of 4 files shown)


freeekanayaka:
I've never used `uv_getaddrinfo`, but I suspect the extra dance in this PR is not needed. If we can't cancel a getaddrinfo request, we simply just discard any result if `t->closing` is non-zero (as in the `uvTcpConnectUvWriteCb` and `uvTcpConnectUvConnectCb` callbacks).

Basically something like this (declaration order inverted and extra comments added just for clarity):

```C
int UvTcpConnect(...) {
    ...
    assert(!t->closing);  /* Note that t->closing is false at the beginning */
    ...
    QUEUE_PUSH(&t->connecting, &r->queue); /* So we know there's a connection attempt in progress, 
                  and we abort it in UvTcpConnectClose() if uvTcpClose() gets
                  called because the engine is shutting down. */.
    ...
    uvTcpConnectStart(r, address);
}

static int uvTcpConnectStart(struct uvTcpConnect *r, const char *address) {
    ...
    rv = uv_tcp_init(r->t->loop, r->tcp); /* we'll need to close this handle in case of failure,
                           see uvTcpConnectAbort */
    assert(rv == 0);
    ...
    rv = uv_getaddrinfo(r->t->loop, &r->getaddrinfo, &uvGetAddrInfoCb, hostname,
                        service, &hints);
    ...
}

static void uvGetAddrInfoCb(uv_getaddrinfo_t *req,
                            int status,
                            struct addrinfo *res) {
    ...
    if (t->closing) { /* This basically checks if uvTcpClose() was called, see uv_tcp.c.
        Note that uvTcpConnectAbort() has been already called too, by UvTcpConnectClose */
        connect->status = RAFT_CANCELED;
        return;
    }

    if (status != 0) { /* The DNS resolution failed, explicitly abort the whole attempt */
        assert(status != UV_ECANCELED); /* t->closing would have been true, because
               we attempt cancelling in uvTcpAbort()  */
        connect->status = RAFT_NOCONNECTION;
        ...
        goto err;
    }

    rv = uv_tcp_connect(&r->connect, r->tcp, (const struct sockaddr *)res->ai_addr
                        uvTcpConnectUvConnectCb);
    if (rv != 0) {
        /* UNTESTED: since parsing succeed, this should fail only because of
         * lack of system resources */
        connect->status = RAFT_NOCONNECTION;
        goto err;
    }

    return;

err:
    uvTcpConnectAbort(connect);
}

static void uvTcpConnectUvConnectCb(struct uv_connect_s *req, int status) {
    ... /* No change should be needed here */
}

static void uvTcpConnectUvWriteCb(struct uv_write_s *write, int status) {
    ... /* No change should be needed here */
}

static void uvTcpConnectFinish(struct uvTcpConnect *connect) {
   ...
    /* The only change needed here should be freeing the addr structure, since we
        don't need it anymore. Note that this function will be always invoked,
        either by uvTcpConnectUvWriteCb or by uvTcpConnectUvCloseCb, so no leaking will happen.  */
   uv_freeaddrinfo(connect->getaddrinfo.addrinfo);
   ...
}

static void uvTcpConnectAbort(struct uvTcpConnect *connect)
{
    QUEUE_REMOVE(&connect->queue);
    QUEUE_PUSH(&connect->t->aborting, &connect->queue);
    uv_cancel((struct uv_req_s *)&connect->getaddrinfo); /* Best effort, it doesn't matter 
          if it actually cancels or not, in any case the uvGetAddrInfoCb callback will
          be invoked if the request is still in-flight  */
    uv_close((struct uv_handle_s *)connect->tcp, uvTcpConnectUvCloseCb);
}

```

freeekanayaka:
Just to be extra clear: the call to `uv_cancel()` is entirely optional, the code should work just fine if we remove it. In fact, we don't even try to cancel the `uv_write()` call that happens after the connect, we just let it finish and check if we possibly need to abort.

NorbertHeusser:
Not canceling the uv_write is totally fine as the uv_write is dependent on the same tcp handle as the tcp_connect. In this case it is fine and totally safe to simply close the tcp handle and all pending or currently processed on this handle will be stopped before the callback give to the uv_close will be invoked. This hold for any of-of-band invocation from a thread outside the libuv loop as well.
The code you proposed was my first attempt as well. But this way we would introduce a race condition, with the libuv library working on invalid (freed) memory. As the getaddrinfo itself (in contrast to network operations based on libepoll) is not really asynchronous it is wrapped by some threads inside the libuv. If the cancellation of the async getaddrinfo fails this may have two reasons: 
1. The getaddrinfo was already performend an the callback was already invoked. We could have introduced an additional member/flag to indicate this, but simply used the state of the tcp handle to detect this case.
2. The getaddrinfo request is currently being processed internally inside the libuv and cannot be canceled anymore. The given callback maybe invoked any time or currently is in execution. 
So if we simply invoke the uv_close on the tcp handle right after we have tried to cancel the getaddrinfo request the libuv will immediately close the tcp handle (as it is not used for anything up to now) and invoke the give cb. And this callback will invoke the uvTcpConnectFinish, which will cleanup all the memory structures. But depending on the current execution stack of the getaddrinfo request in the libuv our getaddrinfo callback was accessing freed memory. Additionally the finish cb will try to free memory (the getaddrinfo result), which may not been allocated.
Happily this was easy to reproduce running very detailed unit test resulting in SEGFAULT and core dump...
This was the reason to sped the additional effort on the extra dance.

NorbertHeusser:
So the uv_cancel is not really necessary to stop the connect, but we need to make sure the structures used in uvGetAddrInfoCb are still valid until the callback execution has finished.

freeekanayaka:
Ah, you're right, I now see what you mean.

What about adding a flag that tracks whether the getaddrinfo request is in-flight or not, basically along the lines of what you had suggested at the very beginning I believe.

Same code as above with a little modification:

```C
struct uvTcpConnect {
    bool resolving; /* will be initialized to false */
}


static int uvTcpConnectStart(struct uvTcpConnect *r, const char *address) {
    ...
    rv = uv_tcp_init(r->t->loop, r->tcp); /* we'll need to close this handle in case of failure,
                           see uvTcpConnectAbort */
    assert(rv == 0);
    ...
    rv = uv_getaddrinfo(r->t->loop, &r->getaddrinfo, &uvGetAddrInfoCb, hostname,
                        service, &hints);
    if (rv != 0) {...}
    r->resolving = true; /* this means that the uv_getaddrinfo call is in progress */
    ...
}

static void uvGetAddrInfoCb(uv_getaddrinfo_t *req,
                            int status,
                            struct addrinfo *res) {
    ...

    connect->resolving = false; /* run this before anything else */

    if (t->closing) { /* This basically checks if uvTcpClose() was called, see uv_tcp.c.
        Note that uvTcpConnectAbort() has been already called too, by UvTcpConnectClose,
        but uv_close() was not called yet because the connect->resolving flag was true */
        connect->status = RAFT_CANCELED;

        /* call uv_close() now that the getaddrinfo request has completed */
        uv_close((struct uv_handle_s *)connect->tcp, uvTcpConnectUvCloseCb);

        return;
    }
    ... /* the rest is the same as above */
}

static void uvTcpConnectAbort(struct uvTcpConnect *connect)
{
    QUEUE_REMOVE(&connect->queue);
    QUEUE_PUSH(&connect->t->aborting, &connect->queue);
    uv_cancel((struct uv_req_s *)&connect->getaddrinfo); /* Best effort, it doesn't matter 
          if it actually cancels or not, in any case the uvGetAddrInfoCb callback will
          be invoked if the request is still in-flight  */

    /* Only call uv_close() if there's no getaddrinfo request in flight. Otherwise, we
        delay it until it is completed. */ 
    if (!connect->resolving) {
        uv_close((struct uv_handle_s *)connect->tcp, uvTcpConnectUvCloseCb);
    }
}

```

NorbertHeusser:
The solution you proposed was one I though about as well. But decided not to use it for several reasons:
1. From the design perspective the code sections checking the t->closing flag are the same in all callbacks. With the proposed solution thios design would have been broken and in behavior would be different for this callback.
2. Second design perspective: Right now the uv_close on the tcp handle is executed either right before everything is handed of to the libuv (due to a failure in the uvTcpConnectStart). Or initiated from the  uvTcpConnectAbort and not from any of the callbacks, which belong to the normal processing flow. This makes the close flow very transparent in my view. The newly introduced check cb with the close is in my perspective directly related to the abort and less harming in transparency.
3. The most important point: With 23 years of experience in implementing multi-threaded solutions I learned it's always hard to make sure to really hit all kind of concurrent execution with the introduction of these kind of check flags. And this kind of race conditions are sometimes hard to detect in unit tests (normally running single threaded). In real concurrent execution it may work well 99,9% of the time and hit you hard, if it crashes in production. e.g in the code you proposed the libuv may be right in the execution of uvGetAddrInfoCb. And the execution stack right after the _if t->closing_ section. So starting at the code section you described with _/* the rest is the same as above */_. The current LWP maybe out aside by the system scheduler for whatever reason. Now the main thread is executing it's uvTcpConnectAbort. The uv_cancel will fail (as the CB is in execution), the _connect->resolving_ will be false and the uv_close will be executed and the given uv_close cb will be invoked immediately by the libuv as the connect on the tcp handle has not been started. All structs wills be removed and viola the continuation of the  uvGetAddrInfoCb will continue to work on stale memory.... 

This was the reason for the solution I proposed. As the current solution in the PR is based on the guarantees given by the libuv.

NorbertHeusser:
Is there a way to retrigger the CLA check as I filed the contributor agreement now ?

MathieuBordere:
> Is there a way to retrigger the CLA check as I filed the contributor agreement now ?

It failed again, possibly because there is some delay between you signing the agreement and the sytem actually knowing that you signed. I'll keep an eye on it if the problem would persist.

freeekanayaka:
> The solution you proposed was one I though about as well. But decided not to use it for several reasons:
> 
> 1. From the design perspective the code sections checking the t->closing flag are the same in all callbacks. With the proposed solution thios design would have been broken and in behavior would be different for this callback.

This is actually a very common design throughout this raft library (and libuv itself). Basically the two callbacks that we have now can be aborted at the same time for the reason you mentioned (because they both depend on the tcp handle). This third callback we are introducing is an independent one that introduces an additional ordering requirement. Using the flag we make sure that ordering is preserved: first we let the uv_getaddrinfo call to finish and the uvGetAddrInfoCb callback to be triggered and only then we proceed with uv_close().

We already do this kind of callback ordering in various places.

> 2. Second design perspective: Right now the uv_close on the tcp handle is executed either right before everything is handed of to the libuv (due to a failure in the uvTcpConnectStart). Or initiated from the  uvTcpConnectAbort and not from any of the callbacks, which belong to the normal processing flow. This makes the close flow very transparent in my view. The newly introduced check cb with the close is in my perspective directly related to the abort and less harming in transparency.

The uv_close() call due to a failure of uvTcpConnectStart is pretty trivial and it's basically just a direct rollback.

The role of uvTcpConnectAbort() is to trigger an orderly shutdown. The fact that we don't call uv_close() right away actually makes things very explicit in my view: there is a shutdown/abort order to honor and at this time we can't yet call uv_close().

> 3. The most important point: With 23 years of experience in implementing multi-threaded solutions I learned it's always hard to make sure to really hit all kind of concurrent execution with the introduction of these kind of check flags. And this kind of race conditions are sometimes hard to detect in unit tests (normally running single threaded). In real concurrent execution it may work well 99,9% of the time and hit you hard, if it crashes in production. e.g in the code you proposed the libuv may be right in the execution of uvGetAddrInfoCb. And the execution stack right after the _if t->closing_ section. So starting at the code section you described with _/* the rest is the same as above */_. The current LWP maybe out aside by the system scheduler for whatever reason. Now the main thread is executing it's uvTcpConnectAbort. The uv_cancel will fail (as the CB is in execution), the _connect->resolving_ will be false and the uv_close will be executed and the given uv_close cb will be invoked immediately by the libuv as the connect on the tcp handle has not been started. All structs wills be removed and viola the continuation of the  uvGetAddrInfoCb will continue to work on stale memory....

This can't happen because of the single-threaded event-driven design of raft/libuv. There is just one thread, and any callback that is invoked by libuv is guaranteed to run from start to end without any other libuv or user callback code being executed, regardless of the OS scheduling. It's very different from multi-threaded designs were I totally agree that flags like this are usually a bad idea, or should at least be protected with very careful mutexes. There are no race conditions of that kind in event-loop based designs.

In this case if libuv is right in the execution of uvGetAddrInfoCb, in particular starting the t->closing section that you mention, then there is no way that the "main thread" executes uvTcpConnectAbort in parallel. There is only one main thread, and it's executing uvGetAddrInfoCb, so nothing else can't execute until the uvGetAddrInfoCb function has returned.

Again, this kind of design pattern is very common in both libraft and libuv, and I believe very solid.

I believe there is actually one subtle problem in the current code of this PR, I'm going to add a separate comment for it. I'd recommend to stick with the `t->resolving` flag approach since it makes ordering explicit and has less moving parts (like the additional `uv_check` handle).

NorbertHeusser:
Got it. As the libraft is totally running single threaded and using the same thread to invoke the uv_run periodically there is no concurrency at all. So will I will replace the uv_check with the flag you prefer and update the PR.

freeekanayaka:
> Got it. As the libraft is totally running single threaded and using the same thread to invoke the uv_run periodically there is no concurrency at all. So will I will replace the uv_check with the flag you prefer and update the PR.

Exactly! Thank you for bearing with me, that's appreciated :)

NorbertHeusser:
Appreciate it as well. Always interested to find a good solution, which fit well into the existing code base and keep the code consistent, stable and maintainable.
Replaced the check now with the bool flag. All tests still ran fine locally.

src/uv_tcp_connect.c
```diff
@@ -146,55 +155,111 @@ static void uvTcpConnectUvConnectCb(struct uv_connect_s *req, int status)
...
 {
-    struct UvTcp *t = r->t;
-    struct sockaddr_in addr;
+    struct uvTcpConnect *connect = req->data;
+    struct UvTcp *t = connect->t;
     int rv;
 
-    rv = uvIpParse(address, &addr);
+    connect->resolving =
+        false; /* Indicate we are in the name resolving phase */
```
freeekanayaka, line 168: Probably better to write "Indicate that we are done with the resolving phase"

freeekanayaka (approved):

freeekanayaka:
Looks good to me, thanks again.

Just in case you are interested, since you mentioned that "there is no concurrency at all", there is actually concurrency in this kind of single-threaded event-based systems. However there is no parallelism, which is different. This is a nice talk about it: https://www.youtube.com/watch?v=tIrVLcUq4xE (it's not really Go-specific).

MathieuBordere:
@NorbertHeusser Is `norbert.heusser@cedalo.com` the email address you signed up with at Launchpad?

src/uv_ip.c
```diff
@@ -7,6 +7,52 @@
...
+        if (!service_ptr) {
+            return RAFT_NAMETOOLONG;
+        }
+    }
+    if (service) {
+        if (!service_ptr) {
+            service_ptr = strchr(address, colon);
+        }
+        if (!service_ptr || *service_ptr == 0 || *(++service_ptr) == 0) {
+            service_ptr = "8080";
```
MathieuBordere, line 47: What's special about `8080`? Maybe document this somewhere?

src/uv_tcp_connect.c
```diff
@@ -25,14 +25,16 @@
...
-    queue queue;                 /* Pending connect queue */
+    struct UvTcp *t;                     /* Transport implementation */
+    struct raft_uv_connect *req;         /* User request */
+    uv_buf_t handshake;                  /* Handshake data */
+    struct uv_tcp_s *tcp;                /* TCP connection socket handle */
+    struct uv_getaddrinfo_s getaddrinfo; /* DNS resolve request */
+    struct uv_connect_s connect;         /* TCP connection request */
+    struct uv_write_s write;             /* TCP handshake request */
+    int status;                          /* Returned to the request callback */
+    bool resolving; /* Indicate name resolving in progress */
```
MathieuBordere, line 36: Aligning with the comments above would be a bit more aesthetically pleasing.

NorbertHeusser:
Yes. This is the e-mail adress I used for the launchpad account as well. 
Signed the contributor agreement now once again. Maybe something went wrong the first time as the launchpad Account was new as well.
Is there a way to see signing the agreement was successfull  somewhere in the launchpad account ?
Did not get any kind of confirmation e-mail about signing the contributor agreement.

src/uv_tcp_connect.c
```diff
@@ -146,55 +155,111 @@ static void uvTcpConnectUvConnectCb(struct uv_connect_s *req, int status)
...
+
+    r->handshake.base = NULL;
+
     /* Initialize the handshake buffer. */
     rv = uvTcpEncodeHandshake(t->id, t->address, &r->handshake);
     if (rv != 0) {
         assert(rv == RAFT_NOMEM);
-        ErrMsgOom(r->t->transport->errmsg);
+        ErrMsgOom(t->transport->errmsg);
         goto err;
```
MathieuBordere, line 221: `handshake.base` is freed after `err`, but never allocated. It looks safe right now, but it's not very pretty and can lead to an erroneous free in the future.

MathieuBordere (changes requested):

src/uv_tcp_connect.c
NorbertHeusser: I initialise the pointer to NULL in the beginning of the function. As the raft_free (like the posix free is safe to be invoked on a NULL pointer) it safe from my perspective.

src/uv_tcp_connect.c
NorbertHeusser: As the project has a .clang-format file in it's root folder I ran a clang-format on all files I touched. My expectation was all format will be fine after that.

src/uv_tcp_connect.c
freeekanayaka: It's just the default port if none is specified. We should indeed document it, or perhaps return an error instead.

PR #303, opened 2022-08-31
API update

MathieuBordere:
Proposal to update the raft API to allow for future extensions without breaking binaries linked against older versions.

Probably easiest to review by commit, it's draft so input is more then welcome.


include/raft.h
```diff
@@ -47,22 +59,22 @@ enum {
...
  */
-typedef unsigned long long raft_term;
+typedef uint64_t raft_term;
 
 /**
- * Hold the value of a raft entry index. Guaranteed to be at least 64-bit long.
+ * Hold the value of a raft entry index.
  */
-typedef unsigned long long raft_index;
+typedef uint64_t raft_index;
```
freeekanayaka, line 72: I'm not entirely convinced by these type changes. They seem to make a difference only if the underlying value exceeds the 64bit threshold, however in that case we just shift the problem because the value will overflow and restart from 0, regardless of the fact that it can be serialized "correctly".

I feel the proper fix would be to check that we don't exceed the 64bit threshold when those values are generated by the lib (in case of raft_term, raft_index, raft_time) or provided by the user (in case of raft_time).

Also, these are somehow abstract types and serialization concerns should be kept separate.

freeekanayaka: Also, in principle a certain architecture could have a "native" hardware type for ```unsigned long long``` that is more efficient than ```uint64_t```. This is of course purely speculative and it most probably wouldn't have a practical impact of any kind (performance or otherwise), but it's just to it's just to illustrate that generic C types are usually a better choice because the compiler has more freedom to map it to the most appropriate hardware type that does the job.

src/log.h
```diff
@@ -8,8 +8,53 @@
...
+    struct                       /* Information about last snapshot, or zero. */
+    {
+        raft_index last_index; /* Snapshot replaces all entries up to here. */
+        raft_term last_term;   /* Term of last index. */
+    } snapshot;
+};
+
 /* Initialize an empty in-memory log of raft entries. */
-void logInit(struct raft_log *l);
+void logInit(struct raft_log **l);
```
cole-miller, line 57: Maybe `struct raft_log *logInit(void);` instead?

cole-miller (commented):
Maybe I'm missing something -- if we're adding a version field to `raft_uv_transport` to prepare for evolving it in the future, why don't we also add a `reserved` field as for some of the other structs?

include/raft/fixture.h
```diff
@@ -70,6 +72,7 @@ struct raft_fixture
     struct raft_fixture_event *event; /* Last event occurred. */
     raft_fixture_event_cb hook;       /* Event callback. */
     struct raft_fixture_server *servers[RAFT_FIXTURE_MAX_SERVERS];
+    uint64_t _reserved[16];           /* For future expansion of struct. */
```
cole-miller, line 75: nit: this has an underscore where the other `reserved` fields don't.

include/raft.h
freeekanayaka: I'll add that in practical terms, `raft_time` will take >500 million years to overflow, same time frame for `raft_index` if we consider a steady flow of one transaction per millisecond. For `raft_term` overflowing the headroom is even higher, and for `raft_id` the `raft_digest()` should be generally used (or some other user-defined method to avoid duplicates) and there too the headroom is even higher as well because configuration changes are even more sparse than term bumps.

All in all, I'd suggest to keep things as they are.

src/log.h
MathieuBordere: indeed, that's a bit more idiomatic.

MathieuBordere:
> Maybe I'm missing something -- if we're adding a version field to `raft_uv_transport` to prepare for evolving it in the future, why don't we also add a `reserved` field as for some of the other structs?

Maybe the difference is somewhat arbitrary, but I consider those structs with reserved space 'internal' structs, meaning that raft uses them for whatever it wants, but they can still be allocated by the user or embedded in another struct (was one of the design decisions), and that's why we need enough room to add functionality.

The versioned structs are 'external', meaning the client fills out their functionality, they too can be allocated by the client or embedded in another struct. Based on the version, as long as the raft library doesn't read/write members that do not correspond to the version, memory accesses should be safe. In theory we could add padding to those structs, but I don't think it is needed imo. 

So for internal structs we don't want the version approach because it will hamper us to add transparent functionality purely because the user did not allocate enough space for it.

include/raft.h
stgraber: The thing I'd like to make sure here is that we don't use types that may change size depending on architecture as if that makes it over the wire or to disk, we may hit some issues.

MathieuBordere: I think for the `raft_time`, `raft_term` and `raft_index` I would leave things as-is then, not even checking overflow as it's so improbable. For the `raft_id` I think it still makes sense to change the type to `uint64_t`, we (hopefully) get free compiler warnings that way if, on some exotic system, the user would ever assign a larger value to a `raft_id`. And we don't have to check for user supplied values that overflow that way.

freeekanayaka: > The thing I'd like to make sure here is that we don't use types that may change size depending on architecture as if that makes it over the wire or to disk, we may hit some issues.

Serialization is indeed a separate concern, as I was mentioning. The `unsigned long long` type is guaranteed to be at least 64bit on any architecture (and the compiler can use the hardware type that is best fit for that). We know that practically speaking we won't overflow those 64bits (and even if that was a possible thing we should add boundary checks instead, regardless). When we serialize those values we always use an architecture-independent encoding with specific sizes.

freeekanayaka: > I think for the `raft_time`, `raft_term` and `raft_index` I would leave things as-is then, not even checking overflow as it's so improbable. For the `raft_id` I think it still makes sense to change the type to `uint64_t`, we (hopefully) get free compiler warnings if, on some exotic system, the user would ever assign a larger value to a `raft_id`. And we don't have to check for user supplied values that overflow.

I'm not sure to understand this. This a runtime concern no? If a user assigns a larger value, it would just overflow and wrap around at run time, with no compiler warnings whatsoever.

If we find that a concern (which at least to me doesn't seem to be the case, practically speaking), I'd suggest to introduce a boundary check. I believe adding that check in `raft_bootstrap()` and `raft_add()` would be enough, since if I'm not mistaking those are the only ways to introduce new IDs.

MathieuBordere: Yeah looks like I was overestimating the C typesystem and that something like:

```C
#include <inttypes.h>

typedef uint8_t raft_id;

static void f(raft_id id)
{
    (void) id;
}

int main(int argc, char **argv)
{
   uint64_t id_too_large = (uint64_t)1 << 63;
   f(id_too_large);
   return 0;
}
```

would let the compiler complain, but it doesn't.

freeekanayaka: > Yeah looks like I was overestimating the C typesystem and that something like:
> 
> ```c
> #include <inttypes.h>
> 
> typedef uint8_t raft_id;
> 
> static void f(raft_id id)
> {
>     (void) id;
> }
> 
> int main(int argc, char **argv)
> {
>    uint64_t id_too_large = (uint64_t)1 << 63;
>    f(id_too_large);
>    return 0;
> }
> ```
> 
> would let the compiler complain, but it doesn't.

Yeah. If you want to introduce a boundary check, actually the best place would not be what I mentioned above but rather the `configurationEncodeToBuf()` function in `configuration.c` where we actually convert the `unsigned long long` to a 64bit little-endian representation. Adding the check there makes sense since this is really a serialization concern from our perspective. In terms of what value the user chooses, we have no control on that, if it overflows `unsigned long long` that's a user concern, not ours. Plus, the recommended way to generate an ID is with `raft_digest()`, which is safe.

include/raft.h
```diff
@@ -336,6 +309,22 @@ enum {
...
+ *
+ * Updating to a new version of an RPC message is done based on the encoded size
+ * of the message, see e.g. `sizeofRequestVoteV1` in uv_encoding.c
+ *
+ * The RPC message types all have a `reserved` field. When extending such a
+ * struct, and thus shrinking the `reserved` array, one should be careful not to
+ * give meaning the `zero` value because the raft implementation needs to know
+ * whether to other side knows the new field or not. `zero` means the other side
+ * does not know the new field/functionality (this is why raft_tribool exists).
+ * Implementers MUST always zero out unknown fields before passing them to raft.
```
freeekanayaka, line 321: I didn't quite realize this when we introduced raft_tribool, but now that I see the `0`-means-older-code-on-the-other-side approach generalized here I'm thinking that perhaps this not the most natural approach.

In most cases `0` is an important and natural value to use. For example let's say that we introduce a new counter field in a message. Do we then have use `0` for "the other side did not include this field", `1` for `0`, `2` for `1`, etc?

Plus, it creates an asymmetry between fields that we already have and the just follow the natural semantics for their values and fields that we add in the future.

Actually we already see the asymmetry in `struct raft_request_vote_result`, where the `vote_granted` field is a regular `bool` and the `pre_vote` field is a `raft_tribool`, which looks weird to me. If we had thought about this pre-vote issue since the very first version of this library, we would certainly have used `bool` not `raft_tribool`, which is something telling. The fact that we use `raft_tribool` now seems a smell that we are mixing two separate concerns: one concern is to store a boolean value in a struct, the other concern is to know whether the other side did include this value in the message it sent us.

This makes me think that a more appropriate approach would be perhaps to separate these two concerns and be explicit about what the other side sent. There are various ways I can imagine to do that, for example adding a `version` field to `struct raft_message` or other variations of that idea. Before we discuss the detail about this possible approach though, I'd like to know how do you feel about this.

MathieuBordere: Yes, I'm certainly open to suggestions. The `zero` approach feels indeed a bit like a 'plumbing' solution.

freeekanayaka: Ok, then my suggestion would be to add an `int version` field to all the various message structures (`struct raft_request_vote`, `struct raft_request_vote_result`,  `struct raft_append_entries`, etc).

Note that I believe this is a bit better than adding the `version` field to `struct raft_message` as I was hinting above, both because what we're versioning is really the specific messages (and not `struct raft_message`), and because as bonus points it should also minimize code changes (since we often pass around the specific messages, not whole `struct raft_message` objects).

This would be pretty much similar to the `version` fields in the `raft_io`/`raft_fsm` structures, making the various message structures effectively "external" structures (borrowing the same terminology that you used in [this comment](https://github.com/canonical/raft/pull/303#issuecomment-1234043078)).

It makes sense considering that those message structures are really part of the `raft_io` interface (specifically the `send()` and `raft_io_recv_cb` parts), so they "inherit" the fact of being external. When receiving messages via `raft_io_recv_cb`, the `raft_io` implementation should be in charge of telling `raft` core what to fields to expect, and it can be done via this new `version` field (like it does for its methods, via the `raft_io->version` field). When sending messages `raft` core should tell `raft_io` what fields it's providing, again via the `version` field.

Regarding encoding and decoding in the libuv-based `raft_io` implementation, we currently use the very first 8 bytes of a message to store the message type (see [here](https://github.com/canonical/raft/blob/09a821f48a827cd75024ad8c689ea7b252ba72b0/src/uv_recv.c#L253) and [here](https://github.com/canonical/raft/blob/09a821f48a827cd75024ad8c689ea7b252ba72b0/src/uv_encoding.c#L213)). I'd suggest to change that slightly, in a fully backward compatible way. We should use the first byte to store the message type (a range of 255 is more than enough even for the future), the second byte to store the version, and the last two bytes are unused and can be reserved for future use. We can decide that version `0` is the current version of all messages, and current servers are already setting the second byte to `0`, so we have backward compatibility.

Note that this also allows us to change the semantics of a message without changing its wire-level length, as at the moment we're using the message length as a "proxy" for its version.

As a side note, with this explicit message versioning approach we should not need to add padding to the various message structures, in the same way we don't use padding for `raft_io`/`raft_fsm`.

To keep things light, in this branch we could just add the `version` field to all the message structures and nothing else. The actual implementation of the encoding and decoding could be done separately. Or we could add the implementation as well, it shouldn't be a lot of code.

MathieuBordere: We could use the first 2 bytes as the message type, the next 2 bytes as version and rest is unused. Then the cast to `unsigned short` [here](https://github.com/canonical/raft/blob/09a821f48a827cd75024ad8c689ea7b252ba72b0/src/uv_encoding.c#L485) should still be fine on old clients with the version set to non-0, or am I missing something?

freeekanayaka: > We could use the first 2 bytes as the message type, the next 2 bytes as version and rest is unused. Then the cast to `unsigned short` [here](https://github.com/canonical/raft/blob/09a821f48a827cd75024ad8c689ea7b252ba72b0/src/uv_encoding.c#L485) should still be fine on old clients with the version set to non-0, or am I missing something?

I'm not sure what you mean. If version is set to non-0, old clients would get a broken message type no? Because they use [`byteFlip64`](https://github.com/canonical/raft/blob/09a821f48a827cd75024ad8c689ea7b252ba72b0/src/uv_recv.c#L253) to get the type.

Unfortunately I believe the only option is to look at the message length, as we already do.

freeekanayaka: We could still modify the current code to use just 2 bytes for the message type, so perhaps down the line we could encode the version in the last 2 bytes, when we are reasonably sure that it's very unlikely to have old clients around.

freeekanayaka: Actually I just realized now what you mean about the cast. I'm not sure the cast helps, since `unsigned short` is at least 16 bits, but it can be longer if the compiler decides to, so no trimming would take place AFAIU.

MathieuBordere: Yeah I assumed it was 2 bytes on all platforms. Practically I think we could do with 4 byte type, 2 byte version and 2 byte reserved/unused, but it would not be portable. I don't think we would ever come across a system with wider than 4-byte unsigned shorts though.

freeekanayaka: > Yeah I assumed it was 2 bytes on all platforms. Practically I think we could do with 4 byte type, 2 byte version and 2 byte reserved/unused, but it would not be portable. I don't think we would ever come across a system with wider than 4-byte unsigned shorts though. I don't think I could get myself to do this, so let's stick with the size approach then.

Agreed.

MathieuBordere: So, practically, we would still version the message structs, but instead of relying on an explicit version in the on wire message, we would rely on the size of the on wire message.

freeekanayaka: > So, practically, we would still version the message structs, but instead of relying on an explicit version in the on wire message, we would rely on the size of the on wire message.

Exactly. As said, I'd also change the code to read only the first 2 bytes to decode the message type, so maybe at some point down the road we'll feel that enough time has passed that it's very unlikely to have both old code and new code servers operating together.

MathieuBordere: > > So, practically, we would still version the message structs, but instead of relying on an explicit version in the on wire message, we would rely on the size of the on wire message.
> 
> Exactly. As said, I'd also change the code to read only the first 2 bytes to decode the message type, so maybe at some point down the road we'll feel that enough time has passed that it's very unlikely to have both old code and new code servers operating together.

I was doing it the other way around, only writing 2 bytes when encoding the message, but that's not strictly needed now, reading only 2 bytes when decoding is what is needed.

freeekanayaka: > > > So, practically, we would still version the message structs, but instead of relying on an explicit version in the on wire message, we would rely on the size of the on wire message.
> > 
> > 
> > Exactly. As said, I'd also change the code to read only the first 2 bytes to decode the message type, so maybe at some point down the road we'll feel that enough time has passed that it's very unlikely to have both old code and new code servers operating together.
> 
> I was doing it the other way around, only writing 2 bytes when encoding the message, but that's not strictly needed now, reading only 2 bytes when decoding is what is needed.

Exactly.

MathieuBordere: >  I'd also change the code to read only the first 2 bytes to decode the message type, so maybe at some point down the road we'll feel that enough time has passed that it's very unlikely to have both old code and new code servers operating together.

I downcast the `uint64_t` type to `uint16_t`, I thought that might be the least intrusive way for now to only use the first 2 bytes of preamble.

freeekanayaka: > > I'd also change the code to read only the first 2 bytes to decode the message type, so maybe at some point down the road we'll feel that enough time has passed that it's very unlikely to have both old code and new code servers operating together.
> 
> I downcast the `uint64_t` type to `uint16_t`, I thought that might be the least intrusive way for now to only use the first 2 bytes of the preamble.

Sounds right to me.

src/uv_encoding.c
```diff
@@ -408,6 +415,7 @@ static int decodeAppendEntries(const uv_buf_t *buf,
 
     cursor = buf->base;
 
+    args->version = 1;
```
freeekanayaka, line 418: Shouldn't we set the initial version to `0`? In this way the wire format what we intend to use in the future is already correct (I'm referring to the idea of encoding the version in the message preamble).

I'm also fine setting the initial version to `1`, I guess we'll just have to remember to convert `0` to `1` whenever we'll switch to the new wire format. I'm just pointing this out to make sure we're not missing something.

include/raft.h
```diff
@@ -336,6 +312,22 @@ enum {
...
  * A single RPC message that can be sent or received over the network.
+ *
+ * The RPC message types all have a `version` and a `reserved` field.
+ * In the libuv io implementation, `version` is filled out during decoding
+ * and is based on the size of the message on the wire, see e.g.
+ * `sizeofRequestVoteV1`. The version number in the RAFT_MESSAGE_XXX_VERSION
+ * macro needs to be bumped every time the message is changed.
+ *
+ * The `reserved` fields are there to leave enough room for future extensions
+ * because messages could be allocated by a client.
```
freeekanayaka, line 323: As far as I understand, the `reserved` field is not needed anymore now. The `version` field acts the same as the one that we use for "external" structures (like `raft_io` and `raft_fsm`).

Anything "producing" (i.e. allocating) a message structure must also set the version field. That way the consumer will know what fields are present and what fields are not present and it won't try to access fields beyond what was allocated.

Furthermore, I probably mentioned that these message structures are effectively part for the `raft_io` interface: they are meant to be used only for communication between `raft` core and `raft_io`. So the only place where they can be allocated is either `raft` core or a `raft_io` implementation. They won't be allocated by "normal" final users of the library.

MathieuBordere:
I'm working on the dqlite disk-mode in parallel, I'm currently figuring out a decent proposal for the snapshot mechanism, and analyze if anything needs to be added to this PR for that.

MathieuBordere:
I think this is ready for review, let's maybe not merge it on a Friday as it would bump the .so number and break builds.

include/raft.h
```diff
@@ -598,6 +596,8 @@ typedef void (*raft_close_cb)(struct raft *raft);
 struct raft_change;   /* Forward declaration */
 struct raft_transfer; /* Forward declaration */
 
+struct raft_log;
+
 /**
  * Hold and drive the state of a single raft server in a cluster.
  */
```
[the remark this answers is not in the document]
MathieuBordere: I thought about it yes, but the argument to leave that as-is is to allow for stack allocation and embedding the `struct raft` in other structs (like dqlite is currently doing)

include/raft.h
```diff
@@ -104,6 +107,7 @@ struct raft_server
 
 /**
  * Hold information about all servers currently part of the cluster.
+ * WARNING: This struct is encoded/decoded, be careful when adapting it.
  */
 struct raft_configuration
```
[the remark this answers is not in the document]
MathieuBordere: I think it could make sens to hide this one, there are already `raft_configuration_init` and `raft_configuration_close` anyway. `dqlite` likes to access the configuration members directly currently so we would need to add a bit of API to facilitate that.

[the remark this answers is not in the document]
freeekanayaka: I think `struct raft_configuration` needs to be public, because it's part of `struct raft_snapshot` which any `raft_io` implementation (potentially third party) must be able to serialize.

include/raft.h
```diff
@@ -201,74 +205,34 @@ struct raft_entry
...
 /**
  * Hold the result of a RequestVote RPC.
  */
 struct raft_request_vote_result
 {
+    int version;
     raft_term term;        /* Receiver's current term (candidate updates itself). */
     bool vote_granted;     /* True means candidate received vote. */
-    raft_tribool pre_vote; /* The response to a pre-vote RequestVote or not. */
+    bool pre_vote;         /* The response to a pre-vote RequestVote or not. */
```
freeekanayaka, line 233: This is a nice change, if I understand correctly the `raft_tribool` type is not needed anymore because we version messages? However, won't there issues with backward compatibility?

I didn't quite look in detail, just asking to conform it's ok.

MathieuBordere:
> I just remembered that for [canonical/dqlite#420](https://github.com/canonical/dqlite/issues/420) we want to support support including a request ID in various places like `struct raft_apply`, I don't think that's addressed by the current branch?
> 
> Edit: and see @freeekanayaka's comment in that issue about client sessions, which might raise other ABI/API concerns
Thanks, I'll take a look!

include/raft.h
MathieuBordere: Backwards compatibility on the wire you mean?

freeekanayaka: > Backwards compatibility on the wire you mean?

Yes, I believe `0` currently has a special meaning (i.e. neither true or false, iow, the other side didn't include it). If we move to regular `bool` that will be `false`. As said, I'm not entirely sure of the details, I was wondering if problems could arise from an old node talking to a new node.

In any case, all things considered, even if there were issue, I think they would be minor and unlikely enough to not bother and keep this change.

MathieuBordere: It should be okay, it's just a different translation to the new struct of a flags argument of `RequestVoteResult` on the wire. In the past if we received the old message we would fill out the struct with `raft_tribool_unknown`, now we fill it out with `version = 1` and ignore the new bool `pre_vote` struct field.

MathieuBordere:
> . . .
> - at raft_start, we initially don't know any config information, so both `configuration` and `known_good_config` are empty
>
> . . .

At `raft_start` we do know the configuration, if there's a snapshot, we know the configuration in the snapshot was committed. If there are no other config entries in succeeding log entries, we fill `configuration` and `known_good_configuration` out with the configuration from the snapshot. If there is a snapshot and 1 config entry in the log, the config entry in the log can be committed or uncommitted, we can still fill out `known_good_config` with the one from the snapshot and `configuration` with the one from the log and set `configuration_uncommitted_index`, waiting for the entry to be committed or rolled back. If there are 2 or more entries in the log succeeding the snapshot, we know that the second to last entry must have been a committed config (as a raft node won't accept a new config change unless the previous one is committed or truncated/rolled back from the log) and we use it as `known_good_configuration` and the last one as `configuration` and set `configuration_uncommitted_index`.

MathieuBordere:
I've added a `req_id` and `reserved` fields to `RAFT__REQUEST` now. I also added `RAFT_REQUEST` to the other requests, `raft_change` and `raft_transfer`, those 2 only use the `data*` field but I thought it was nicer to make the requests uniform. Let me know if you don't like the change.

MathieuBordere:
Should I bump https://github.com/canonical/raft/blob/fde5378a44ad5d7ed2225f5beaabdb20af230746/configure.ac#L2 too? I'm inclined to do so (without making a release though).

src/request.h
```diff
@@ -11,6 +11,8 @@ struct request
     int type;
     raft_index index;
     void *queue[2];
+    uint64_t id;
```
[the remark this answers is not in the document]
MathieuBordere: Yeah, forgot to adapt it after a first iteration. I'll refactor a bit so we don't have to think about keeping those 2 files in sync.

MathieuBordere:
So I've added 3 id's to `RAFT__REQUEST` now and changed the type to a 16 byte array to accommodate UUID's.

MathieuBordere:
I propose to merge this next Monday, unless @freeekanayaka has any objections.

freeekanayaka (approved):

Issue #304, opened 2022-09-01
can't restart when I try to apply string message to cluster


cagev:
I'm  new to raft , I modified the example server to apply a string buffer to cluster. 
if start the server directly,  It seems to work , the followers can print the applied message from leader. 
but when I stop the leader server, and restart it , it can't be started, with the errors: 

./bin/msgserver ./data1 1 
1: starting
1: raft_start(): io: load closed segment 0000000000000002-0000000000000005: entries batch 2 starting at byte 63: read header: short read: 153 bytes instead of 6272
1: stopping

my code is  here 

https://github.com/cagev/raftdemo/blob/main/demoserver.cpp
line 328 - 338 , I apply a string to cluster. 

I really can't figure out what's the problem is , so post here to get more help. 
sorry to bother !

MathieuBordere:
My gut feeling is that it's related to this TODO https://github.com/canonical/raft/blob/09a821f48a827cd75024ad8c689ea7b252ba72b0/src/uv_segment.c#L756 i.e. your buffer length needs to be a multiple of 8.

PR #305, opened 2022-09-02
Uvappend 8-byte alignment

MathieuBordere:
Should prevent issues like: https://github.com/canonical/raft/issues/304

### DIFF
--- a/src/uv.h
+++ b/src/uv.h
@@ -318,7 +318,8 @@ int UvPrepare(struct uv *uv,
  * and then remove it immediately. */
 void UvPrepareClose(struct uv *uv);
 
-/* Implementation of raft_io->append.*/
+/* Implementation of raft_io->append. All the raft_buffers of the raft_entry structs
+ * in the entries array are required to have a len that is a multiple of 8. */
 int UvAppend(struct raft_io *io,
              struct raft_io_append *req,
              const struct raft_entry entries[],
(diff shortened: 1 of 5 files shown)


src/uv.h
```diff
@@ -318,7 +318,8 @@ int UvPrepare(struct uv *uv,
  * and then remove it immediately. */
 void UvPrepareClose(struct uv *uv);
 
-/* Implementation of raft_io->append.*/
+/* Implementation of raft_io->append. All the raft_buffers of the raft_entry structs
+ * in the entries array are required to have a len that is a multiple of 8. */
```
cole-miller, line 322: Do we want the alignment requirement to be part of the contract of raft_io->append more generally?

MathieuBordere: I don't think so, it's an implementation detail of the uv backend imo.

src/uv_append.c
```diff
@@ -622,6 +622,24 @@ static int uvAppendEnqueueRequest(struct uv *uv, struct uvAppend *append)
     return rv;
 }
 
+/* Ensure all entry buffers are 8-byte aligned */
+static int uvEnsureEntryBuffersAligned(struct uv *uv,
```
freeekanayaka, line 626: Nit: I'd rename this to `uvCheckEntryBuffersAligned`. Throughout the code base "ensure" usually means things like "see if this buffer is at least N bytes, and enlarge it if it's not" (i.e. "make sure things are in a certain way"), while "check" usually means "return an error if something is not in a certain way".

PR #306, opened 2022-09-02
test/integration/append_helpers: Fix unaligned access

cole-miller:
#305 introduced a test with unaligned log entries, but the ENTRIES macro in test/integration/append_helpers.h can generate an unaligned access in this situation.

Should fix the latest Launchpad build failures.


MathieuBordere (approved):
Looks good

PR #307, opened 2022-09-06
Travis bump

MathieuBordere:
Bump Travis Ubuntu version from 18.04 -> 20.04


MathieuBordere:
Thought it would have been a trivial bump, but it seems to give issues. Will probably move away from travis by using [run-on-architecture](https://github.com/marketplace/actions/run-on-architecture).

Issue #298, opened 2022-08-16
Add support for hostnames instead of static IP addresses


NorbertHeusser:
We would like to use the raft library in the context of a docker swarm runtime environment. In contrast to traditional docker runtime it's not possible to assign internal static IP addresses to the individual cluster members. Static IP assignment not working seems to be an open issue in docker swarm now for several year so I don't expect it to be get fixed soon.

I tried to implement the DNS lookup in our application layer and keep the raft library config on IP addresses. But this has several caveats: 
- The name service in a docker swarm environment for an instance is added once the instance has started. So performing the name resolution to early will fail.
- If the instance gets moved to a different worker node the dynamic IP address of this instance may changed and I cannot change it in the raft config.

I guess in general it will make sense to support hostnames in the raft config instead of static IPs. And I already start to add the necessary code extension to provide a pull request. But I would need some design guidance how you would prefer the solution to be later on accepted into you code base.

The base problem about adding host name resolving is the DNS request will add an additional and potentially blocking call to the TCP connection handling. The libuv used in your integration implementation provides an async API to perform the name resolving asynchronously. In general I currently see to code paths, where a name resolve would need to happen:
1. In the outgoing TCP connection to the other cluster members. This code path is already implemented asynchronous and I would need to add an additional asynchronous step before the connect itself.
2.Before the bind of the listen port for the address (or the bind_address). The current code path does not include any asynchronous steps for the bind/list calls. So in this case I would implement the name resolving in a traditional potential timeout blocking getaddrinfo base call. Adding the name resolving as an asynchronous step would need some general design changes in the bind/listen handling.

The design questions, which are currently open to me are:
- Are you fine adding a synchronous and potentially timeout blocking name resolving to the bind/listen socket implementation code path ?
- As the name resolving will add an addition asynchronous step to the outgoing tcp connection I may encapsulate it totally insite the uv_tcp_listen.c implementation. This way I would need to add an additional internal state flag telling me wether the connection is still in name resolving or already trying to connect. Or I would need to add a new queue into the uv_tcp.h/.c (e.g. name_resolving) and a single tcp-connection would be added to this queue first and moved to the connecting queue once the name resolving is finished. I personally would prefer the first way encapsulating it inside the uv_tcp_connect, but would respect your decision, if you prefer an additional queue to be more transparent.

freeekanayaka:
Thanks for looking at this! It'd be a nice feature to have.

Regarding implementation:

[...]

PR #309, opened 2022-09-08
Feature hostnames

NorbertHeusser:
This is the second PR to finalize hostname support for the raft library. This will close https://github.com/canonical/raft/issues/298

### DIFF
--- a/src/uv_ip.c
+++ b/src/uv_ip.c
@@ -32,7 +32,7 @@ int uvIpAddrSplit(const char *address,
 {
     char colon = ':';
     const char *service_ptr = NULL;
-
+    
     if (host) {
         service_ptr = strCpyUntil(host, address, host_size, colon);
         if (!service_ptr) {
@@ -53,29 +53,33 @@ int uvIpAddrSplit(const char *address,
     return 0;
 }
 
-int uvIpParse(const char *address, struct sockaddr_in *addr)
+/* Synchronoues resolve hostname to IP address */
+int uvIpResolveBindAddresses(const char *address, struct addrinfo **ai_result)
 {
-    char buf[256];
-    size_t n;
-    char *host;
-    char *port;
-    char *colon = ":";
+    static struct addrinfo hints = {
+        .ai_flags = AI_V4MAPPED | AI_ADDRCONFIG | AI_PASSIVE | AI_NUMERICSERV,
+        .ai_family = AF_INET,
+        .ai_socktype = SOCK_STREAM,
+        .ai_protocol = 0};
+    char hostname[NI_MAXHOST];
+    char service[NI_MAXSERV];
     int rv;
-
-    /* TODO: turn this poor man parsing into proper one */
-    n = sizeof(buf) - 1;
-    strncpy(buf, address, n);
-    buf[n] = '\0';
-    host = strtok(buf, colon);
-    port = strtok(NULL, ":");
-    if (port == NULL) {
-        port = "8080";
-    }
-
-    rv = uv_ip4_addr(host, atoi(port), addr);
+    
+    rv = uvIpAddrSplit( address, hostname, sizeof(hostname), service, sizeof(service));
     if (rv != 0) {
-        return RAFT_NOCONNECTION;
+        return rv;
     }
-
+    
+    if (hostname[0]) {
+        rv = getaddrinfo( hostname, service, &hints, ai_result);
+    } else {
+        rv = getaddrinfo( NULL, service, &hints, ai_result);
+    }
+    
+    if (rv != 0) {
+        return RAFT_IOERR;
+    }
+    
     return 0;
 }
+
--- a/src/uv_ip.c
+++ b/src/uv_ip.c
@@ -57,7 +57,7 @@ int uvIpAddrSplit(const char *address,
 int uvIpResolveBindAddresses(const char *address, struct addrinfo **ai_result)
 {
     static struct addrinfo hints = {
-        .ai_flags = AI_V4MAPPED | AI_ADDRCONFIG | AI_PASSIVE | AI_NUMERICSERV,
+        .ai_flags = AI_ADDRCONFIG | AI_PASSIVE | AI_NUMERICSERV,
         .ai_family = AF_INET,
         .ai_socktype = SOCK_STREAM,
         .ai_protocol = 0};
--- a/src/uv_ip.c
+++ b/src/uv_ip.c
@@ -32,7 +32,7 @@ int uvIpAddrSplit(const char *address,
 {
     char colon = ':';
     const char *service_ptr = NULL;
-    
+
     if (host) {
         service_ptr = strCpyUntil(host, address, host_size, colon);
         if (!service_ptr) {
@@ -64,22 +64,22 @@ int uvIpResolveBindAddresses(const char *address, struct addrinfo **ai_result)
     char hostname[NI_MAXHOST];
     char service[NI_MAXSERV];
     int rv;
-    
-    rv = uvIpAddrSplit( address, hostname, sizeof(hostname), service, sizeof(service));
+
+    rv = uvIpAddrSplit(address, hostname, sizeof(hostname), service,
+                       sizeof(service));
     if (rv != 0) {
         return rv;
     }
-    
+
     if (hostname[0]) {
-        rv = getaddrinfo( hostname, service, &hints, ai_result);
+        rv = getaddrinfo(hostname, service, &hints, ai_result);
     } else {
-        rv = getaddrinfo( NULL, service, &hints, ai_result);
+        rv = getaddrinfo(NULL, service, &hints, ai_result);
     }
-    
+
     if (rv != 0) {
         return RAFT_IOERR;
     }
-    
+
     return 0;
 }
-
--- a/src/uv_tcp.c
+++ b/src/uv_tcp.c
@@ -20,11 +20,11 @@ static int uvTcpInit(struct raft_uv_transport *transport,
     assert(address != NULL);
     t->id = id;
     t->address = address;
-    rv = uv_tcp_init(t->loop, &t->listener);
+    rv = uv_tcp_init(t->loop, &t->default_listener);
     if (rv != 0) {
         return rv;
     }
-    t->listener.data = t;
+    t->default_listener.data = t;
     return 0;
 }
 
@@ -49,7 +49,7 @@ void UvTcpMaybeFireCloseCb(struct UvTcp *t)
     assert(QUEUE_IS_EMPTY(&t->accepting));
     assert(QUEUE_IS_EMPTY(&t->connecting));
 
-    if (t->listener.data != NULL) {
+    if (t->listeners || t->default_listener.data != NULL) {
         return;
     }
     if (!QUEUE_IS_EMPTY(&t->aborting)) {
@@ -78,7 +78,9 @@ int raft_uv_tcp_init(struct raft_uv_transport *transport,
     t->id = 0;
     t->address = NULL;
     t->bind_address = NULL;
-    t->listener.data = NULL;
+    t->default_listener.data = NULL;
+    t->listeners = NULL;
+    t->num_listeners = 0;
     t->accept_cb = NULL;
     QUEUE_INIT(&t->accepting);
     QUEUE_INIT(&t->connecting);
@@ -106,10 +108,12 @@ int raft_uv_tcp_set_bind_address(struct raft_uv_transport *transport,
                                  const char *address)
 {
     struct UvTcp *t = transport->impl;
-    struct sockaddr_in addr;
+    char hostname[NI_MAXHOST];
+    char service[NI_MAXSERV];
     int rv;
 
-    rv = uvIpParse(address, &addr);
+    rv = uvIpAddrSplit(address, hostname, sizeof(hostname), service,
+                       sizeof(service));
     if (rv != 0) {
         return RAFT_INVALID;
     }
--- a/src/uv_tcp.c
+++ b/src/uv_tcp.c
@@ -74,7 +74,7 @@ int raft_uv_tcp_init(struct raft_uv_transport *transport,
     t->address = NULL;
     t->bind_address = NULL;
     t->listeners = NULL;
-    t->num_listeners = 0;
+    t->n_listeners = 0;
     t->accept_cb = NULL;
     QUEUE_INIT(&t->accepting);
     QUEUE_INIT(&t->connecting);
--- a/src/uv_tcp.c
+++ b/src/uv_tcp.c
@@ -47,7 +47,7 @@ void UvTcpMaybeFireCloseCb(struct UvTcp *t)
         return;
     }
 
-    if (t->listeners) {
+    if (t->listeners != NULL) {
         return;
     }
 
--- a/src/uv_tcp.c
+++ b/src/uv_tcp.c
@@ -1,5 +1,5 @@
-#include "uv_ip.h"
 #include "uv_tcp.h"
+#include "uv_ip.h"
 
 #include <string.h>
 
--- a/src/uv_tcp.c
+++ b/src/uv_tcp.c
@@ -32,6 +32,7 @@ static void uvTcpClose(struct raft_uv_transport *transport,
     t->close_cb = cb;
     UvTcpListenClose(t);
     UvTcpConnectClose(t);
+    UvTcpMaybeFireCloseCb(t);
 }
 
 void UvTcpMaybeFireCloseCb(struct UvTcp *t)
@@ -40,12 +41,13 @@ void UvTcpMaybeFireCloseCb(struct UvTcp *t)
         return;
     }
 
-    if (t->listeners) {
+    assert(QUEUE_IS_EMPTY(&t->accepting));
+    assert(QUEUE_IS_EMPTY(&t->connecting));
+    if (!QUEUE_IS_EMPTY(&t->aborting)) {
         return;
     }
-    assert(QUEUE_IS_EMPTY(&t->accepting));
 
-    if (!QUEUE_IS_EMPTY(&t->connecting) || !QUEUE_IS_EMPTY(&t->aborting)) {
+    if (t->listeners) {
         return;
     }
 
--- a/src/uv_tcp.c
+++ b/src/uv_tcp.c
@@ -15,16 +15,10 @@ static int uvTcpInit(struct raft_uv_transport *transport,
                      const char *address)
 {
     struct UvTcp *t = transport->impl;
-    int rv;
     assert(id > 0);
     assert(address != NULL);
     t->id = id;
     t->address = address;
-    rv = uv_tcp_init(t->loop, &t->default_listener);
-    if (rv != 0) {
-        return rv;
-    }
-    t->default_listener.data = t;
     return 0;
 }
 
@@ -46,13 +40,12 @@ void UvTcpMaybeFireCloseCb(struct UvTcp *t)
         return;
     }
 
-    assert(QUEUE_IS_EMPTY(&t->accepting));
-    assert(QUEUE_IS_EMPTY(&t->connecting));
-
-    if (t->listeners || t->default_listener.data != NULL) {
+    if (t->listeners) {
         return;
     }
-    if (!QUEUE_IS_EMPTY(&t->aborting)) {
+    assert(QUEUE_IS_EMPTY(&t->accepting));
+
+    if (!QUEUE_IS_EMPTY(&t->connecting) || !QUEUE_IS_EMPTY(&t->aborting)) {
         return;
     }
 
@@ -78,7 +71,6 @@ int raft_uv_tcp_init(struct raft_uv_transport *transport,
     t->id = 0;
     t->address = NULL;
     t->bind_address = NULL;
-    t->default_listener.data = NULL;
     t->listeners = NULL;
     t->num_listeners = 0;
     t->accept_cb = NULL;
--- a/src/uv_tcp.h
+++ b/src/uv_tcp.h
@@ -13,7 +13,9 @@ struct UvTcp
     struct uv_loop_s *loop;              /* Event loop */
     raft_id id;                          /* ID of this raft server */
     const char *address;                 /* Address of this raft server */
-    struct uv_tcp_s listener;            /* Listening TCP socket handle */
+    struct uv_tcp_s default_listener;    /* The default listener to use on single IP */
+    size_t num_listeners;                /* The number of listener sockets (multi IPs) */
+    struct uv_tcp_s* listeners;          /* The array of tcp listen sockets (multi IPs) */
     raft_uv_accept_cb accept_cb;         /* Call after accepting a connection */
     queue accepting;                     /* Connections being accepted */
     queue connecting;                    /* Pending connection requests */
--- a/src/uv_tcp.h
+++ b/src/uv_tcp.h
@@ -13,7 +13,6 @@ struct UvTcp
     struct uv_loop_s *loop;              /* Event loop */
     raft_id id;                          /* ID of this raft server */
     const char *address;                 /* Address of this raft server */
-    struct uv_tcp_s default_listener;    /* The default listener to use on single IP */
     size_t num_listeners;                /* The number of listener sockets (multi IPs) */
     struct uv_tcp_s* listeners;          /* The array of tcp listen sockets (multi IPs) */
     raft_uv_accept_cb accept_cb;         /* Call after accepting a connection */
--- a/src/uv_tcp.h
+++ b/src/uv_tcp.h
@@ -13,7 +13,7 @@ struct UvTcp
     struct uv_loop_s *loop;              /* Event loop */
     raft_id id;                          /* ID of this raft server */
     const char *address;                 /* Address of this raft server */
-    size_t num_listeners;                /* The number of listener sockets (multi IPs) */
+    unsigned n_listeners;                /* The number of listener sockets (multi IPs) */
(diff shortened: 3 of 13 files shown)


NorbertHeusser:
I don't know, why the builds are currently failing on 18.04. And suggestions welcome.
Will try to reproduce it locally tomorrow, but need to setup a ubuntu 18.04 for this.

NorbertHeusser:
Did an additional force push with just a whitespace formatting correction to force a new cla-check.

MathieuBordere:
> OK. Fixed all build issues now. And signed all commits.
> 
> Re-signed the contributor agreement the third time now :-(

The cla-check finally recognized you ... hooray for that :-) Will review soon.

NorbertHeusser:
I am happy the cla-check works now for :-)

Perhaps a few comments on the design. As the getaddrinfo may result into multiple IP addresses I need to deal with that in the different parts of the solution. This holds for the synchronous system call as well as for the asynchronous lbuv based invocation. The solution design I have chosen are:
- For the outgoing connection I need to try one IP after the other until I find a first one working. This will cover the case one of the IP addresses is not open for listening. Once I am able to connect to the port I expect the handshake to work as well. Another solution would have been to try the next IP additionally in case the handshake is failing. But I did not consider this to be a common error scenario. As I need to close the tcp handle after each attempt and reopen it I added an additional bool flag retry to react on it in the abort (otherwise we could try to close the handle twice). It would be possible to combine the two flags resolving and retry in a single one as never both will be set at the same time, but I find the current solution better maintainable.
- For the listen socket I have to bind all the resolved IP addresses. But to my experience I have seen a bind address resolving to multiple IP addresses very rarely. For this reason I have chosen to keep the current dedicated listener and rename it to default_listener as this single listener will be used in most cases. As I know after the name resolve wether I need multiple listeners I will create multiple listener only in case they are needed and in this rare case don't use the default_listener. This ways it's easier to close everything in a clear way as well. 
- Due to an old glibc behavior to reply with duplicate entries  (which is fixed in some linux distros like SLES)  I added a small workaround for this as well. Otherwise I would end up in an address already in use error.
- For testing purpose I implement a kind of getaadrinfo/freeaddrinfo mock. Into the testing library. The behavior of this helper file is described in the header. To avoid the need to provide all getaddrinfo mock responses for all test cases and to make sure everything works with the real system getaddrinfo the mock function will forward the calls to the real implementation, if it is not enabled. And the mock functionality is used in the tests cases which want to test some special cases like multiple IPs only.

src/uv_ip.c
```diff
@@ -32,7 +32,7 @@ int uvIpAddrSplit(const char *address,
 {
     char colon = ':';
     const char *service_ptr = NULL;
-
+
```
freeekanayaka, line 35: Spurious blanks here.

freeekanayaka:
Looks overall good to me at first sight. Before I dig deeper: how much important is it to support multiple listeners? Could we instead return an error if the hostname resolves to multiple IPs? Or choose the first IP.

I know that's sub-optimal, just wondering if it is worth the cost of additional complexity or if it's something that will probably never be used in practice. For example is it relevant for your own use case?

Since you already did all the work it's a pity to simplify it, but if there's no immediate need for multiple listeners we could keep this code around in a branch and only introduce this support when there's a request for it.

I guess same reasoning for outgoing connections.

NorbertHeusser:
Regarding the listen sockets dealing with multiple IP addresses immediately comes into play, if we would introduce handling of IPv6 addresses. As systems with IPv6 only are still extremely rare today most systems have dual mode using IPv4 and IPV6 at the same time. So with activation of IPv6 additionally to IPv4 most hostnames will resolve one (or more) IPv6 addresses and an IPv4 address.

For outgoing TCP connections the most common use case I have seen in the past is the configuration of multiple network interfaces and IP addresses to achieve fail safety. And this is beside the cluster scale-out  concept one of the most common use cases to install a cluster as a HA solution. This might be less common over the last year due to the cheaper availability of network components supporting network bonding.
 
Right now I need the hostname feature to be able to run our cluster in a docker swarm environment. And this does not support static IPs. But the customer plans to use multiple different virtual networks over his docker swarm. So I donÄt know right now, how fast we need it. But sooner or later we will have a customer asking for it.

And in general the I don't like implementing half of a solution (was already thinking about adding IPv6 support straight away as well). Hostnames resolving to multiple IPs is nothing new, but has been the case since the beginning of the IP protocol definition. And to my opinion dealing with this should be default implementation in all IP network base applications.

freeekanayaka:
Fair enough, thanks for explaining. I'll give a deeper look at the PR.

test/lib/addrinfo.c
```diff
@@ -0,0 +1,167 @@
...
+    struct addrinfo_mock_data* next;
+};
+
+static struct addrinfo_mock_data* addrinfo_data;
+
+void AddrinfoInjectSetUp(MUNIT_UNUSED const MunitParameter params[])
+{
+    munit_assert_int(addrinfo_mock_enabled, ==, false);
+    munit_assert_ptr((void*)addrinfo_data, ==, NULL);
+    addrinfo_mock_enabled = true;
```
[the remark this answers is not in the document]
NorbertHeusser: When I remember correctly the munit test framework is using fork to separate the individual test case executions. This way the tests should not interfere. But will double check and get back.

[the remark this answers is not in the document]
NorbertHeusser: Double checked it. The muint test framework uses a fork to isolate each test case run using a separate child process. Therefore the addrinfo_mock will have no problem with that.
But parallel execution of all tests using listening ports and tcp connections might be a problem, if two tests executed in parallel use the same port.

test/lib/addrinfo.h
```diff
@@ -0,0 +1,36 @@
+/* Support for getaddrinfo injection for test purpose 
+ *
+ * Provide a local bound version to capture teh getaddrinfo/freeaddrinfo incovation
```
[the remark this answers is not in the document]
NorbertHeusser: Fixed.

src/uv_tcp_connect.c
```diff
@@ -10,11 +10,20 @@
 /* The happy path of a connection request is:
  *
  * - Create a TCP handle and submit a TCP connect request.
+ * - Initiate a asynchronous dns resolve request
```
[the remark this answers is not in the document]
NorbertHeusser: fixed.

src/uv_tcp_connect.c
```diff
@@ -155,22 +202,37 @@ static void uvTcpConnectUvConnectCb(struct uv_connect_s *req, int status)
     uvTcpConnectAbort(connect);
 }
 
+/* Helper function to connect to the remote node */
+static void uvTcpAsyncConnect(struct uvTcpConnect *connect)
+{
+    int rv = uv_tcp_connect(&connect->connect, connect->tcp,
+                            connect->ai_current->ai_addr,
+														uvTcpConnectUvConnectCb);
```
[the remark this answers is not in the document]
NorbertHeusser: Fixed.

src/uv_ip.c
NorbertHeusser: fixed.

NorbertHeusser:
Fixed the requested typos/errors in comments. And applied a clang-format (except for struct comments) to align the formatting.

freeekanayaka:
Looks good overall.

The main bit I'm not entirely convinced about is the differentiation between `default_listener` and `listeners`. In my mind it'd be simpler to actually have just `listeners`, because in any case you have to handle all the intricacies of the multiple-listeners case no matter what. It's true that `listeners` will have only 1 item most of the times, but that's fine, the logic will stay the same.

NorbertHeusser:
> The main bit I'm not entirely convinced about is the differentiation between `default_listener` and `listeners`. In my mind it'd be simpler to actually have just `listeners`, because in any case you have to handle all the intricacies of the multiple-listeners case no matter what. It's true that `listeners` will have only 1 item most of the times, but that's fine, the logic will stay the same.

What you describe was the first attempt I tried to implement, but I ran into problems with the close/destruction order of some corner cases, when I tried to do this. Will try to describe what I observed. The root cause, if I manage the listeners totally dynamically the listeners array (even for a single listener) will be created after the successful response from the name resolving. In case the name resolving does not work we will end up without any listener tcp handle. But in this case we don't have any handle to close in the UvTcpListenClose and therefor don't trigger the uvTcpListenCloseCbListener. And therefore the uv_tcp_listen will not trigger the UvTcpMaybeFireCloseCb. This all work fine as long as there are outgoing connection, which where connected successfully. In this case the outgoing connections will trigger the UvTcpMaybeFireCloseCb and all structures will get removed. But without any outgoing connections the structures will not be freed. I tried to solve this by direct invocation of UvTcpMaybeFireCloseCb from the UvTcpListenClose in case the listener array is not created. But this will fail, if there are outgoing connections, because they expect the whole tcp object to exists until their close_cb functions are finished. 

More generally description:
So there is a destructions dependency between the tcp_listen and tcp_connect part of the code, which work fine right now as both code parts will always have at least have one handle to close. So the code flow will allways close all handles. All close-CB will be invoced and the last one will finally trigger the destruction of the tcp object grouping both. 
This clear order of close/destruction no longer works , if one source code does not have a handle to close any more.

freeekanayaka:
I see, thanks. I'll have to look more closely myself as well then, perhaps there's a way out.

freeekanayaka:
Ok, after reading the code I understand the situation better. I believe we should handle this ordering issue with the same pattern that we use throughout the code: queues (which is what libuv itself does, and basically most asynchronous programs, in way form or another).

My suggestion would be

1. Replace the `struct uv_tcp_s listener;` field of `struct UvTcp` with a `queue listeners` queue object. This queue will contain all active listeners (see the next point).
2. Add a new `struct uvTcpListener` to `uv_tcp_listen.c`, it should look roughly like this:

```C
/* Hold information about a single listening socket. */
struct uvTcpListener
{
    struct UvTcp *t;                 /* Transport implementation */
    struct uv_tcp_s tcp;             /* Listening TCP socket handle */
    queue queue;                     /* Current queue (either listeners or aborting) */
};
```

3. Modify `UvTcpListen` to create an `uvTcpListener` object for each required address, initialize its `tcp` object and call `uv_tcp_bind`/`uv_listen` on it, and finally insert it in the `listeners` queue. This is basically a slight modification of what you do in `UvTcpListenOnMultipleIP`
4. Modify `UvTcpListenClose` to start closing all active listeners, something like:

```C
void UvTcpListenClose(struct UvTcp *t)
{
    queue *head;
    assert(t->closing);
    assert(!QUEUE_IS_EMPTY(&t->listeners));

    while (!QUEUE_IS_EMPTY(&t->accepting)) {
        struct uvTcpIncoming *incoming;
        head = QUEUE_HEAD(&t->accepting);
        incoming = QUEUE_DATA(head, struct uvTcpIncoming, queue);
        uvTcpIncomingAbort(incoming);
    }

    while (!QUEUE_IS_EMPTY(&t->listeners)) {
        struct uvTcpListener *listener;
        head = QUEUE_HEAD(&t->listener);
        listener = QUEUE_DATA(head, struct uvTcpListener, queue);
        uvTcpListenerAbort(listener);
    }
}
```

5. Implement `uvTcpListenerAbort` with something like:

```C
/* The listening TCP handle has been closed, release all memory
 * associated with the listener object. */
static void uvTcpListenerCloseCb(struct uv_handle_s *handle)
{
    struct uvTcpListener *listener = handle->data;
    struct UvTcp *t = listener->t;
    QUEUE_REMOVE(&listener->queue);
    RaftHeapFree(listener);
    UvTcpMaybeFireCloseCb(t);
}

/* Stop listening for incoming connection on the given listener. */
static void uvTcpListenerAbort(struct uvTcpListener *listener)
{
    struct UvTcp *t = listener->t;
    /* After uv_close() returns we are guaranteed that no more listen callback
     * will be called. */
    QUEUE_REMOVE(&listener->queue);
    QUEUE_PUSH(&t->aborting, &listener->queue);
    uv_close((struct uv_handle_s *)&listener->tcp, uvTcpListenerCloseCb);
}
```

6. Modify `UvTcpMaybeFireCloseCb` so it checks only the aborting queue and not `t->listener.data` (which does not exist anymore):

```C
void UvTcpMaybeFireCloseCb(struct UvTcp *t)
{
    if (!t->closing) {
        return;
    }

    assert(QUEUE_IS_EMPTY(&t->listeners));
    assert(QUEUE_IS_EMPTY(&t->accepting));
    assert(QUEUE_IS_EMPTY(&t->connecting));

    if (!QUEUE_IS_EMPTY(&t->aborting)) {
        return;
    }

    if (t->close_cb != NULL) {
        t->close_cb(t->transport);
    }
}
```

I believe that should work neatly, unless I'm missing something.

One important note to add is that in the case you mentioned, where the name resolving does not work for the listening address, then the `UvTcpListen` function should just return an error and the closing sequence shouldn't even be triggered (essentially this error will bubble up and `raft_start()` will return it, before the event loop is even started).

I'm even not entirely sure why you raised the issue of the name resolving not working, because as I said in that case things are simple, it's just error propagation before the loop is even started, so no actual closing sequence is involved (and certainly there won't be any outgoing connections attempts ongoing).

Hope it all makes sense.

NorbertHeusser:
After reading your long comment about adding an additional queue I was not really sure, if this is really what we need. 
The problem I was facing was not about the UvTcpMaybeFireCloseCb being invoked to ealry, but instead nto being invoked at all. Especially your comments about everything around listen is synchronous and should not be a problem let me get back one more time and check, if the problems I was facing were about the unit tests only.

So I removed the default listener from the code, which made the code around the listener much more straight forward. To avoid the UvTcpMaybeFireCloseCb not being invoked at all I now invoke it directly from UvTcpListenClose in case there is no listener. 
I got it all working by replacing a assert in the UvTcpMaybeFireCloseCb with an if check (as the UvTcpMaybeFireCloseCb maybe invoked one more time now). And remove one assertion from the unit test (as the tcp close cb maybe invoked in some unit test now directly from the close).

Please take a look, if the PR maybe accepted this way.

src/uv_tcp_listen.c
```diff
@@ -328,5 +413,12 @@ void UvTcpListenClose(struct UvTcp *t)
         uvTcpIncomingAbort(incoming);
     }
 
-    uv_close((struct uv_handle_s *)&t->listener, uvTcpListenCloseCbListener);
+    if (t->num_listeners) {
+        for (size_t i = 0; i < t->num_listeners; ++i) {
+            uv_close((struct uv_handle_s *)&t->listeners[i],
+                     uvTcpListenCloseCbListener);
+        }
+    } else {
```
freeekanayaka, line 421: Shouldn't we remove this else branch and run an `assert(t->num_listeners > 0)` instead? I don't think there's a case where we have zero listeners (`UvListen` would have returned an error in the first place I think).

src/uv_tcp_listen.c
```diff
@@ -275,51 +285,126 @@ static void uvTcpListenCb(struct uv_stream_s *stream, int status)
     assert(rv != 0);
 }
 
+/* Do bind/listen call on the tcp handle */
+static int UvTcpBindListen(struct uv_tcp_s *listener, struct sockaddr *addr)
```
freeekanayaka, line 289: Nit: should be callsed `uvTcpBindListen`, capital case is only used for exported symbols, not for module-level static ones.

src/uv_tcp_listen.c
```diff
@@ -275,51 +285,126 @@ static void uvTcpListenCb(struct uv_stream_s *stream, int status)
...
+{
+    if (uv_tcp_bind(listener, addr, 0) ||
+        uv_listen((uv_stream_t *)listener, 1, uvTcpListenCb)) {
+        return RAFT_IOERR;
+    }
+    return 0;
+}
+
+/* Create a tcp handle and do bind/listen for each IP */
+static int UvTcpListenOnMultipleIP(struct raft_uv_transport *transport,
```
freeekanayaka, line 299: Same: should be `uvTcpListenOnMultipleIP`

freeekanayaka:
This approach looks good to me, I've left a few minor comments, I'll give it another pass later.

test/integration/test_uv_tcp_listen.c
```diff
@@ -57,28 +57,19 @@ static void acceptCb(struct raft_uv_transport *t,
...
-        _rv = f->transport.init(&f->transport, 1, "127.0.0.1:9000"); \
-        munit_assert_int(_rv, ==, 0);                                \
-        f->transport.data = f;                                       \
-        f->closed = false;                                           \
+#define INIT                                                              \
+    do {                                                                  \
+        int _rv;                                                          \
+        _rv = raft_uv_tcp_init(&f->transport, &f->loop);                  \
+        munit_assert_int(_rv, ==, 0);                                     \
+        const char *bind_addr = munit_parameters_get(params, "b");        \
```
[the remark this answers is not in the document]
NorbertHeusser: Done. replaced parameter name "a" by "address" and "b" by "bind-address".

test/integration/test_uv_tcp_listen.c
```diff
@@ -57,28 +57,19 @@ static void acceptCb(struct raft_uv_transport *t,
...
+        if (bind_addr && strlen(bind_addr)) {                             \
+            _rv = raft_uv_tcp_set_bind_address(&f->transport, bind_addr); \
+            munit_assert_int(_rv, ==, 0);                                 \
+        }                                                                 \
+        _rv = f->transport.init(&f->transport, 1, "127.0.0.1:9000");      \
```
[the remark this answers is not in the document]
NorbertHeusser: Cannot find this line. As it is marked as outdated it seems to be fixed already.

src/uv_tcp_listen.c
NorbertHeusser: Was not aware of this convention. Fixed it for all newly added helper function in uv_tcp_listen.c

src/uv_tcp_listen.c
NorbertHeusser: If I remove the else block a lot of the integration unit tests (e.g. all tcp_connect, append, write, listen,...) will fail as the close_cb will not be invoked.

src/uv_tcp_listen.c
NorbertHeusser: If I remove the else block a lot of the integration unit tests (e.g. all tcp_connect, append, write, listen,...) will fail as the close_cb will not be invoked.

src/uv_tcp_listen.c
NorbertHeusser: Was not aware of this convention. Fixed it for all newly added helper function in uv_tcp_listen.c

src/uv_tcp.c
```diff
@@ -46,13 +40,12 @@ void UvTcpMaybeFireCloseCb(struct UvTcp *t)
         return;
     }
 
-    assert(QUEUE_IS_EMPTY(&t->accepting));
-    assert(QUEUE_IS_EMPTY(&t->connecting));
```
freeekanayaka, line 50: The removal of these assertions looks odd. The `UvTcpMaybeFireCloseCb` is meant to be called only after  `uvTcpClose` is called, which in turn calls `UvTcpListenClose` and `UvTcpConnectClose`. At that point all incoming/outgoing connections in the `accepting`/`connecting` queues should have been moved to the `aborting` queue.

If that's not the case anymore we should understand why.

src/uv_tcp.c
```diff
@@ -46,13 +40,12 @@ void UvTcpMaybeFireCloseCb(struct UvTcp *t)
...
-
-    if (t->listener.data != NULL) {
+    if (t->listeners) {
         return;
     }
-    if (!QUEUE_IS_EMPTY(&t->aborting)) {
+    assert(QUEUE_IS_EMPTY(&t->accepting));
+
+    if (!QUEUE_IS_EMPTY(&t->connecting) || !QUEUE_IS_EMPTY(&t->aborting)) {
```
freeekanayaka, line 48: As said above, the need to check `t->connecting` sounds odd, because it's expected to be empty. Maybe it's a leftover from the previous version with `default_listener`?

Or more probably because the explicit call to `UvTcpMaybeFireCloseCb(t)` in the `else` branch below, where there are no listeners. Since `uvTcpClose` calls `UvTcpListenClose` **before** `UvTcpConnectClose` the `UvTcpMaybeFireCloseCb` callback is called before `UvTcpConnectClose` has been to run, which should not happen.

If that is true, one solution would be to add the `UvTcpMaybeFireCloseCb()` call to then end of `uvTcpClose` instead of `UvTcpListenClose`:

```C
static void uvTcpClose(struct raft_uv_transport *transport,
                       raft_uv_transport_close_cb cb)
{
    struct UvTcp *t = transport->impl;
    assert(!t->closing);
    t->closing = true;
    t->close_cb = cb;
    UvTcpListenClose(t);
    UvTcpConnectClose(t);
    /* Immediately invoke the close callback if there's no
     * asynchronous process to wait for. */
    UvTcpMaybeFireCloseCb(t);
}
```

src/uv_tcp.h
```diff
@@ -13,7 +13,8 @@ struct UvTcp
     struct uv_loop_s *loop;              /* Event loop */
     raft_id id;                          /* ID of this raft server */
     const char *address;                 /* Address of this raft server */
-    struct uv_tcp_s listener;            /* Listening TCP socket handle */
+    size_t num_listeners;                /* The number of listener sockets (multi IPs) */
```
freeekanayaka, line 16: Nit: the type/naming convention we have elsewhere for this kind of counters is `unsigned n_listeners` (so `n_<name-of-array>`).

We're using `unsigned` because we're not really holding the size of a memory object and `unsigned` is big enough for our purposes (it's guaranteed to be at least 16 bits, so up to 65535).

src/uv_tcp_listen.c
freeekanayaka: Please let me know if the above suggestion to put the `UvTcpMaybeFireCloseCb(t)` call at the end of `uvTcpClose` instead solves this issue and the `else` branch can be removed.

freeekanayaka (commented):
Another round of comments.

src/uv_tcp.c
freeekanayaka: > The first assertion only got moved down a few files and is still in place. The later one got replaced by extending the if condition a few lines below. I will explain the need for this replacement in my response a few lines below.

Is the reason the addition of the `vTcpMaybeFireCloseCb(t)` call at then end of `UvTcpListenClose()`? Like I mentioned in the other comment. I'd expect to be just that, or is there more?

src/uv_tcp.c
NorbertHeusser: You are totally right I had to replace the assertion by the if due the direct invocation UvTcpMaybeFireCloseCb. As I wrote in a different comment this was required to get all the unit test working. And we always come back to the same problem. Up to now the tcp-listen part of the tcp object **always** contained a tcp handler, which always got closed. As I removed this tcp handle and replaced it by a dynamic created list of tcp handles there is no longer a guaranteed invocation of the UvTcpMaybeFireCloseCb and a lot of the unit tests will fail. This will never happen in any real usage of the raft library as the init will return an error in this case.
I have no problem to move the UvTcpMaybeFireCloseCb, but in this case I still have to replace the assertion by the if as with this movement the UvTcpMaybeFireCloseCb will **always** be invoked synchronously, even if there are listen sockets opened. But at least the execution order for unit test execution and real usage of the library.

freeekanayaka: I think that as long as the `UvTcpMaybeFireCloseCb` call is at the end of `uvTcpClose`, after `UvTcpListenClose` and `UvTcpConnectClose`, then all assertions should work unchanged. Unless I'm missing something.

src/uv_tcp_listen.c
NorbertHeusser: Moved the UvTcpMaybeFireCloseCb call at the end of the uvTcpClose instead now. And this way it works fine now. This way I could put back in the assertions into the UvTcpMaybeFireCloseCb as well.

src/uv_tcp.c
NorbertHeusser: Moved the UvTcpMaybeFireCloseCb call at the end of the uvTcpClose instead now. And this way it works fine now. This way I could put back in the assertions into the UvTcpMaybeFireCloseCb as well.

src/uv_tcp.h
NorbertHeusser: Renamed the member to n_listeners and changed the type to unsigned.

src/uv_tcp.c
```diff
@@ -48,11 +43,11 @@ void UvTcpMaybeFireCloseCb(struct UvTcp *t)
...
     assert(QUEUE_IS_EMPTY(&t->accepting));
     assert(QUEUE_IS_EMPTY(&t->connecting));
-
-    if (t->listener.data != NULL) {
+    if (!QUEUE_IS_EMPTY(&t->aborting)) {
         return;
     }
-    if (!QUEUE_IS_EMPTY(&t->aborting)) {
+
+    if (t->listeners) {
```
freeekanayaka, line 50: Nit: we usually write this explicitly: ```if (t->listeners != NULL)```, just to help casual readers to know this is a pointer and not a counter or something else.

freeekanayaka (approved):
I've added just one small nit.

Looks good to me, thanks for your work and for bearing with review feedback :)

src/uv_tcp.c
NorbertHeusser: Fixed this as well.

freeekanayaka:
@stgraber @MathieuBordere this can be merged as far as I'm concerned.

MathieuBordere:
Thanks a lot for your contribution!

PR #311, opened 2022-09-15
state: Add getters for a few additional bits of state

cole-miller:
These are used by https://github.com/canonical/dqlite/pull/402 -- see that PR for motivation.


freeekanayaka:
At first sight it feels we're starting to require too much care from the client, which should probably not be concerned with these details.

I don't have a better solution right now, but I'll try to think a bit about it, including the possibility of unconditionally running the barrier at leader election as you originally implemented. This configuration issue is something I hadn't think of.

PR #316, opened 2022-10-04
Add function calls for request lifecycle tracing

cole-miller:
This introduces two trivial functions, `lifecycleRequestStart` and `lifecycleRequestEnd`, that are called when a request is put onto and taken off of the main request queue, respectively. By attaching bpftrace probes to these function calls, we can gather information about how long requests typically spend sitting on this queue. The function definitions are confined to a separate compilation unit to prevent inlining.

We also add an ID field to the request structure so that dequeue and enqueue events for the same request can be correlated; this is a purely internal thing and should be replaced with client-supplied request IDs when we get around to implementing that feature next cycle.


MathieuBordere:
We have to watch out a bit, e.g. `struct raft_apply` contains a `RAFT__REQUEST` and that's embedded [here](https://github.com/canonical/dqlite/blob/7e95f9ff582d1715736d36bfe14f1bad095b57dd/src/leader.h#L28) in dqlite. We would break dqlite libraries compiled against older versions of libraft, I think it needs to be taken into account in the [raft API PR](https://github.com/canonical/raft/pull/303) too.

cole-miller:
Thanks, that's a good point. I'll try removing the ID field and keying the bpftrace map on the other data that's already in the request struct; hopefully that should give enough uniqueness to work with until we're ready to implement real request IDs.

MathieuBordere (approved):

PR #321, opened 2022-10-31
replication: Give up leadership when a disk write fails

### LINKED ISSUE #314: Assertion failure in progressShouldReplicate
I noticed what looks like a significant assertion failure from raft while looking at the logs for a CI run of my jepsen.dqlite branch (https://github.com/canonical/jepsen.dqlite/pull/30). It's this one:

https://github.com/canonical/raft/blob/2b5c68b3755104410906608a391805d3adbafa11/src/progress.c#L116-L118

Here's the log: [app.log](https://github.com/canonical/raft/files/9636452/app.log). The CI job in question is [this one](https://github.com/canonical/jepsen.dqlite/actions/runs/3112903624/jobs/5046862594).

### LINKED ISSUE #320: Partially-replicated entries after a disk write failure
Consider this sequence of events:

1. A raft leader gets an append request. It starts a disk write for the new entry X and, concurrently, replicates X to its followers.
2. A follower receives the entry X and appends it to its log.
3. The leader's disk write fails, and it truncates its log, erasing X.
4. Later, the same leader (in the same term) gets another append request. It appends an entry Y at the same index that X previously occupied and replicates it.
5. The same follower receives the entry Y. Since it already has an entry with the same index/term, it responds with success to the leader *without* updating its log to replace X with Y.
6. Now, the leader can commit Y, but the follower in question will apply X to its state machine instead -- bad!

To fix this, either leaders need to wait until disk writes succeed to replicate the corresponding entries (as I believe the whitepaper envisions), or followers need to replace log entries more aggressively.

cole-miller:
Closes #320
Closes #314 (confirmed by running Jepsen against this branch)

### DIFF
--- a/src/fixture.c
+++ b/src/fixture.c
@@ -224,6 +224,13 @@ static void ioFlushAppend(struct io *s, struct append *append)
 {
     struct raft_entry *entries;
     unsigned i;
+    int status = 0;
+
+    /* Simulates a disk write failure. */
+    if (ioFaultTick(s)) {
+        status = RAFT_IOERR;
+        goto done;
+    }
 
     /* Allocate an array for the old entries plus the new ones. */
     entries = raft_realloc(s->entries, (s->n + append->n) * sizeof *s->entries);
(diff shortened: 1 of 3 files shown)


MathieuBordere:
A test that covers this would be nice to have!

MathieuBordere (approved):

freeekanayaka (approved):

PR #327, opened 2022-11-16
byte: Don't bother with aligned byte-access functions

### LINKED ISSUE #326: Review/replace use of byte access functions that assume alignment
Follow-up to #325. Are there more places where the alignment assumption is not justified? My instinct is to not have the alignment-assuming versions at all, unless there's a strong performance motivation.

cole-miller:
Closes #326

### DIFF
--- a/src/byte.h
+++ b/src/byte.h
@@ -83,19 +83,14 @@ BYTE__INLINE void bytePut8(void **cursor, uint8_t value)
 
 BYTE__INLINE void bytePut32(void **cursor, uint32_t value)
 {
-    uint32_t **p = (uint32_t **)cursor;
-    **p = byteFlip32(value);
-    *p += 1;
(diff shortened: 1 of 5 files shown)


src/byte.h
```diff
@@ -83,19 +83,14 @@ BYTE__INLINE void bytePut8(void **cursor, uint8_t value)
 
 BYTE__INLINE void bytePut32(void **cursor, uint32_t value)
 {
-    uint32_t **p = (uint32_t **)cursor;
-    **p = byteFlip32(value);
-    *p += 1;
+    unsigned i;
+    uint64_t flipped = byteFlip32(value);
```
MathieuBordere, line 87: type should be `uint32_t`

MathieuBordere (approved):

PR #330, opened 2022-11-29
Support printing backtrace when an assertion fails

cole-miller:
More than once while looking at an assertion failure in Jepsen logs I've wished that it came with a backtrace -- this PR implements that.

Specifically, if RAFT_ASSERT_WITH_BACKTRACE is defined during compilation, we call into [libbacktrace](https://github.com/ianlancetaylor/libbacktrace) to print a symbolic backtrace before aborting. I still need to add the build system stuff to enable/disable this feature, find libbacktrace, and pass the appropriate flags during compilation.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -222,6 +222,11 @@ endif # LZ4_ENABLED
 
 endif # UV_ENABLED
 
+if BACKTRACE_ENABLED
+libraft_la_CFLAGS += -DRAFT_ASSERT_WITH_BACKTRACE
+libraft_la_LDFLAGS += -lbacktrace
+endif # BACKTRACE_ENABLED
+
 if EXAMPLE_ENABLED
 
 bin_PROGRAMS += \
--- a/src/assert.h
+++ b/src/assert.h
@@ -19,6 +19,20 @@ extern void munit_errorf_ex(const char *filename,
     do {                 \
         (void)sizeof(x); \
     } while (0)
+#elif defined(RAFT_ASSERT_WITH_BACKTRACE)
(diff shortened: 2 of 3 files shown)


src/assert.h
```diff
@@ -19,6 +19,19 @@ extern void munit_errorf_ex(const char *filename,
     do {                 \
         (void)sizeof(x); \
     } while (0)
+#elif defined(RAFT_ASSERT_WITH_BACKTRACE)
+#include <stdio.h>
+#include <backtrace.h>
+void __assert_fail(const char *, const char *, unsigned, const char *);
```
MathieuBordere, line 25: Silly question maybe, but why do we need to declare this function here?

MathieuBordere: Original approach is fine for me.

Makefile.am
```diff
@@ -222,6 +222,11 @@ endif # LZ4_ENABLED
 
 endif # UV_ENABLED
 
+if BACKTRACE_ENABLED
+libraft_la_CFLAGS += -DRAFT_ASSERT_WITH_BACKTRACE
+libraft_la_LDFLAGS += -lbacktrace
```
MathieuBordere, line 227: Compilation will fail if `libbacktrace` is not found, probably a bit cleaner to detect it during the configure step itself, the `libuv` in `configure.ac` is probably the clearest.

MathieuBordere: Fair enough, I can live with a failure further down the road, it's a debugging feature anyway.

MathieuBordere (approved):

PR #335, opened 2022-12-13
replication: Increase last_applied before firing cb

MathieuBordere:
The `raft_apply` request callback can contain logic that depends on the value of `last_applied`, which logically should have increased after applying the log entry, but before calling the callback.

This can happen e.g. when a barrier entry is applied, and the callback triggers another barrier because it thinks last_applied is running behind the last_index while in reality the log entry has been applied.

### DIFF
--- a/src/replication.c
+++ b/src/replication.c
@@ -1354,6 +1354,9 @@ static int applyCommand(struct raft *r,
     if (rv != 0) {
         return rv;
     }
+
+    r->last_applied = index;
+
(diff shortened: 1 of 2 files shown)


MathieuBordere:
I'll add this commit to the quorum bugfix PR as it's needed there, wanted to logically separate the 2 fixes, but the quorum bugfix actually needs this fix, so I think they belong together.

Issue #339, opened 2022-12-14
Bad memory access (?) in uvClientInit


cole-miller:
This comes from the same Jepsen run as #338, so see that issue for the test tarball. There's a core dump on n4 that seems to indicate a bad memory access inside uvClientInit, with the backtrace looking like this:

```
#6  __strlen_sse2 () at ../sysdeps/x86_64/multiarch/strlen-vec.S:126
#7  0x00007f01d55f6f1f in uvClientInit (address=0x100 <error: Cannot access memory at address 0x100>, id=139644755084192, 
    uv=0x29b1fb0, c=0x7f0194235ef0) at src/uv_send.c:93
#8  uvGetClient (client=<synthetic pointer>, address=0x100 <error: Cannot access memory at address 0x100>, id=139644755084192, 
    uv=0x29b1fb0) at src/uv_send.c:436
#9  UvSend (io=<optimized out>, req=<optimized out>, message=<optimized out>, cb=<optimized out>) at src/uv_send.c:482
#10 0x00007f01d55e306d in sendAppendEntriesResult (r=r@entry=0x29b3ba0, result=result@entry=0x7f019f7fa770) at src/replication.c:817
#11 0x00007f01d55e38c8 in installSnapshotCb (req=<optimized out>, status=<optimized out>) at src/replication.c:1250
#12 0x00007f01d55f7a7b in uvSnapshotPutFinish (put=<optimized out>) at src/uv_snapshot.c:569
#13 uvSnapshotPutAfterWorkCb (work=<optimized out>, status=<optimized out>) at src/uv_snapshot.c:578
```

https://github.com/canonical/raft/blob/02ae98558b32feaf041315c14e1ade14bdc8cdc2/src/uv_send.c#L93

I haven't been able to correlate this with the node's log yet. Looks like a bad value for that `address` parameter.

MathieuBordere:
Yeah it looks like a follower installing a snapshot turns into a candidate? before the snapshot finishes installing.

PR #342, opened 2022-12-15
replication: Only send AppendEntriesResult when FOLLOWER

MathieuBordere:
Should already fix https://github.com/canonical/raft/issues/339

Follow up in https://github.com/canonical/raft/issues/343

### DIFF
--- a/src/replication.c
+++ b/src/replication.c
@@ -1246,7 +1246,10 @@ discard:
     raft_configuration_close(&snapshot->configuration);
 
 respond:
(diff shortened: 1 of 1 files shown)


MathieuBordere:
> I'm a little worried about a potential "ABA problem" where the node makes a whole trip follower->candidate->follower or follower->candidate->leader->follower before installSnapshotCb is invoked -- in that case, we'd use the wrong follower_state. But this seems fine as a stopgap until we figure out #343.

valid remark indeed.

Issue #341, opened 2022-12-15
Double free in ioCloseCb


cole-miller:
See the log here: https://github.com/canonical/dqlite/actions/runs/3700630188/jobs/6269233919. Seems that ioCloseCb is getting invoked twice for the same `struct raft`.

https://github.com/canonical/raft/blob/faadebbe8fae5f7faf1f015b90fae094b3fec7e8/src/raft.c#L110

calvin2021y:
maybe you can try:

1) add log when you start call uvMaybeFireCloseCb, and before return.

pedroleaoc:
I will close this one for now due to lack of a reproducer.

PR #347, opened 2022-12-16
Prevent calling uv->close_cb twice, take 2

cole-miller:
The previous attempt was problematic because `uv->close_cb(uv->io)` could free `uv`, and indeed dqlite's raftCloseCb does this. I've confirmed that dqlite's test suite passes under ASan with this patch.

closes https://github.com/canonical/raft/issues/341

### DIFF
--- a/src/uv.c
+++ b/src/uv.c
@@ -156,6 +156,7 @@ static int uvStart(struct raft_io *io,
 
 void uvMaybeFireCloseCb(struct uv *uv)
 {
+    raft_io_close_cb close_cb;
     tracef("uv maybe fire close cb");
     if (!uv->closing) {
         return;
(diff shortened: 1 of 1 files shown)


MathieuBordere:
> The previous attempt was problematic because `uv->close_cb(uv->io)` could free `uv`, and indeed dqlite's raftCloseCb does this. I've confirmed that dqlite's test suite passes under ASan with this patch.
> 
> Signed-off-by: Cole Miller [cole.miller@canonical.com](mailto:cole.miller@canonical.com)

Do you know where `close_cb` is called twice from? If dqlite's `close_cb` can free `uv` it still seems problematic, I'm not sure if setting `uv->close_cb` to NULL will fix this as dereferencing `uv` is not valid after uv is free?

PR #351, opened 2023-01-04
replication: Drop append entries result if a state transition has occurred

cole-miller:
This should avoid some potential problems when r->follower_state.current_leader changes between receiving a request and calling sendAppendEntriesResult (e.g. the latest manifestation of #339).

### DIFF
--- a/src/replication.c
+++ b/src/replication.c
@@ -845,21 +845,22 @@ static void appendFollowerCb(struct raft_io_append *req, int status)
     assert(args->entries != NULL);
     assert(args->n_entries > 0);
 
-    result.term = r->current_term;
-    result.version = RAFT_APPEND_ENTRIES_RESULT_VERSION;
-    if (status != 0) {
-        if (r->state != RAFT_FOLLOWER) {
-            tracef("local server is not follower -> ignore I/O failure");
-            goto out;
-        }
-        result.rejected = args->prev_log_index + 1;
-        goto respond;
-    }
-
     /* If we're shutting down or have errored, ignore the result. */
     if (r->state == RAFT_UNAVAILABLE) {
         tracef("local server is unavailable -> ignore I/O result");
         goto out;
+    }
+    /* If a new term has started since the request came in, ignore the result. */
+    if (args->term != r->current_term) {
+        tracef("new term while processing request -> ignore I/O result");
+        goto out;
+    }
+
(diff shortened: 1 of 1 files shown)


MathieuBordere:
It's maybe a bit simpler to check if the term has changed in between receiving the request and replying to it?

src/replication.c
```diff
@@ -828,6 +828,7 @@ struct appendFollower
     raft_index index;  /* Index of the first entry in the request. */
     struct raft_append_entries args;
     struct raft_io_append req;
+    raft_term term;    /* Used to check for state changes. */
```
MathieuBordere, line 831: The `struct raft_io_append` above already contains the term of the request, so this is not needed I think.

src/replication.c
```diff
@@ -1206,9 +1211,15 @@ static void installSnapshotCb(struct raft_io_snapshot_put *req, int status)
...
-    /* If we are shutting down, let's discard the result. TODO: what about other
-     * states? */
+    /* If we are shutting down, let's discard the result. */
     if (r->state == RAFT_UNAVAILABLE) {
+        tracef("shutting down -> discard result of snapshot installation");
+        goto discard;
+    }
+    /* Also discard the result if we've transitioned to candidate state since
+     * the request came in. */
+    if (request->term != r->current_term) {
```
MathieuBordere, line 1221: Will probably no longer be needed if #353 is merged?

src/replication.c
```diff
@@ -845,22 +846,24 @@ static void appendFollowerCb(struct raft_io_append *req, int status)
...
+        tracef("state transition while processing request -> ignore I/O result");
+        goto out;
+    }
+
+    result.version = RAFT_APPEND_ENTRIES_RESULT_VERSION;
+    result.term = request->term;
+    if (status != 0) {
+        result.rejected = args->prev_log_index + 1;
+        goto respond;
```
MathieuBordere, line 865: I think we can respond here after converting to `RAFT_UNAVAILABLE` which is unwanted.

src/replication.c
MathieuBordere: sorry, I meant the `struct raft_append_entries`

src/replication.c
MathieuBordere: yeah, sorry, looked over that one.

PR #353, opened 2023-01-04
tick: Don't convert to candidate while a snapshot is installing

### LINKED ISSUE #343: Investigate RAFT_FOLLOWER moving to RAFT_CANDIDATE during snapshot installation.
In https://github.com/canonical/raft/issues/339 it looks like a `RAFT_FOLLOWER` can move to a new state != `RAFT_UNAVAILABLE`, i.e.  `RAFT_CANDIDATE` (`RAFT_LEADER` first goes through candidacy). Investigate if it should be possible to become `RAFT_CANDIDATE` while installing a snapshot. Probably not.

cole-miller:
Closes #343. @MathieuBordere, let me know if this wasn't what you had in mind, I'm sure there are other possible approaches. 

### DIFF
--- a/test/integration/test_snapshot.c
+++ b/test/integration/test_snapshot.c
@@ -717,3 +717,39 @@ TEST(snapshot, takeSnapshotFail, setUp, tearDown, 0, fsm_snapshot_async_params)
     /* No crash or leaks have occurred */
     return MUNIT_OK;
 }
+
+/* A follower doesn't convert to candidate state while it's installing a snapshot. */
+TEST(snapshot, snapshotBlocksCandidate, setUp, tearDown, 0, NULL)
+{
+    struct fixture *f = data;
+    (void)params;
+
+    /* Set very low threshold and trailing entries number */
+    SET_SNAPSHOT_THRESHOLD(3);
+    SET_SNAPSHOT_TRAILING(1);
+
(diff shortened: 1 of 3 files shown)


MathieuBordere:
> I'll try to add a test for this.

Looks good to me, a test would be nice indeed, you could mess with the IO latency and election timeout in the test to hit this case.

test/integration/test_snapshot.c
```diff
@@ -717,3 +717,42 @@ TEST(snapshot, takeSnapshotFail, setUp, tearDown, 0, fsm_snapshot_async_params)
...
+    CLUSTER_DESATURATE_BOTHWAYS(0, 2);
+    fprintf(stderr, "desaturating\n");
+
+    /* Wait a while and check that the leader has sent a snapshot */
+    CLUSTER_STEP_UNTIL_ELAPSED(500);
+    munit_assert_int(CLUSTER_N_SEND(0, RAFT_IO_INSTALL_SNAPSHOT), ==, 1);
+    munit_assert_int(CLUSTER_N_RECV(2, RAFT_IO_INSTALL_SNAPSHOT), ==, 1);
+
+    /* Disconnect the servers again so that heartbeats, etc. won't arrive */
+    fprintf(stderr, "resaturating\n");
```
MathieuBordere, line 751: stray `fprintf`

PR #354, opened 2023-01-05
Config rollback

MathieuBordere:
This PR introduces the following changes:

* Set `configuration_uncommitted_index` upon startup as we can't know if the configuration is committed or not.
* Backup the configuration in case the configuration entry is no longer present in the log.

  1. After taking a snapshot and before purging entries from the log.
  2. After installing a snapshot in case the next config needs to be rolled back.
* Only store committed configurations in a snapshot and not the latest available configuration.

Fixes https://github.com/canonical/raft/issues/250

### DIFF
--- a/include/raft.h
+++ b/include/raft.h
@@ -634,7 +634,8 @@ struct raft
      * 1. #configuration_index and #configuration_uncommitted_index are both
      *    zero. This should only happen when a brand new server starts joining a
      *    cluster and is waiting to receive log entries from the current
-     *    leader. In this case #configuration must be empty and have no servers.
+     *    leader. In this case #configuration and #configuration_previous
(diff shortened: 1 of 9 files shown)


MathieuBordere:
Looks to be misbehaving in Jepsen suite with a lot of `extra online spare` errors. Investigating if it's a fluke or something real.

MathieuBordere:
Jepsen looks fine now. In short, this PR introduces the following changes:

* Set `configuration_uncommitted_index` upon startup as we can't know if the configuration is committed or not.
* Backup the configuration in case the configuration entry is no longer present in the log.

  1. After taking a snapshot and before purging entries from the log.
  2. After installing a snapshot in case the next config needs to be rolled back.

PR #357, opened 2023-01-09
replication: Fix handling of new term when snapshotting

### LINKED ISSUE #356: Assertion `request->term == r->current_term' failed.
Observed in Jepsen:

https://github.com/canonical/jepsen.dqlite/actions/runs/3864574649/jobs/6587461915

cole-miller:
Closes #356 

### DIFF
--- a/src/recv_request_vote.c
+++ b/src/recv_request_vote.c
@@ -3,6 +3,7 @@
 #include "assert.h"
 #include "election.h"
 #include "recv.h"
+#include "replication.h"
 #include "tracing.h"
 
 #define tracef(...) Tracef(r->tracer, __VA_ARGS__)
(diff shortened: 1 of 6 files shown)


src/recv_request_vote.c
```diff
@@ -75,6 +76,12 @@ int recvRequestVote(struct raft *r,
         }
     }
 
+    /* Reject the request if we are installing a snapshot. */
+    if (replicationInstallSnapshotBusy(r)) {
+        tracef("installing snapshot -> reject");
```
MathieuBordere, line 81: This code is not hit in a test, you can verify it by putting an `assert(false)` here.

src/replication.c
```diff
@@ -1214,19 +1215,26 @@ static void installSnapshotCb(struct raft_io_snapshot_put *req, int status)
...
     }
-    /* We don't transition to candidate state while a snapshot is being installed. */
-    assert(request->term == r->current_term);
+    /* If the request is from a previous term, it means that someone else became
+     * a candidate while we were installing the snapshot. In that case, we want to
+     * install the snapshot anyway, but our "current leader" may no longer be the
+     * same as the server that sent the install request, so we shouldn't send a
+     * response to that server. */
+    if (request->term != r->current_term) {
+        tracef("new term since receiving snapshot -> install but don't respond");
```
MathieuBordere, line 1232: This code is also not hit in the test.

MathieuBordere (commented):
I think the approach looks fine, we just need to tweak (or add to) the tests so that the new code paths are hit.

src/recv_request_vote.c
MathieuBordere: Yeah, that's fine, it looks safe.

MathieuBordere (approved):

PR #361, opened 2023-01-10
byte: Tweak signature of byteSha1Transform

cole-miller:
To avoid a (spurious) compiler warning on ppc64el.


MathieuBordere:
I'll close this one as commented in #360

PR #363, opened 2023-01-11
byte: Don't inline byteSha1Update

MathieuBordere:
Inlining seems to cause a false-positive stringop-overread warning on ppc64le, as this is not performance sensitive code in real-world systems, it's okay to not inline this function.


MathieuBordere:
Test failure is unrelated I think, see https://github.com/canonical/raft/issues/364

MathieuBordere:
Will merge to have some feedback on the Launchpad build.

PR #367, opened 2023-01-11
replication: Don't respond to AppendEntries if term has changed

### LINKED ISSUE #339: Bad memory access (?) in uvClientInit

cole-miller:
Another swing at fixing bad accesses to the `current_leader.address` field when a follower updates its term number while entries are being written to disk. This version is significantly simpler than the flawed #351, and I've added a test to cover the new code. Running Jepsen against this branch at https://github.com/canonical/jepsen.dqlite/pull/72 and it looks good so far.

Closes #339 

### DIFF
--- a/src/replication.c
+++ b/src/replication.c
@@ -866,7 +866,7 @@ static void appendFollowerCb(struct raft_io_append *req, int status)
         goto out;
     }
 
-    /* We received an InstallSnapshot RCP while these entries were being
+    /* We received an InstallSnapshot RPC while these entries were being
      * persisted to disk */
     if (replicationInstallSnapshotBusy(r)) {
         goto out;
@@ -911,8 +911,11 @@ static void appendFollowerCb(struct raft_io_append *req, int status)
         }
     }
 
-    if (r->state != RAFT_FOLLOWER) {
-        tracef("local server is not follower -> don't send result");
+    /* If our state or term number has changed since receiving these entries,
+     * our current_leader may have changed as well, so don't send a response
+     * to that server. */
+    if (r->state != RAFT_FOLLOWER || r->current_term != args->term) {
+        tracef("new role or term since receiving entries -> don't respond");
         goto out;
     }
 
--- a/test/integration/test_replication.c
+++ b/test/integration/test_replication.c
@@ -794,6 +794,7 @@ TEST(replication, recvMatch_last_snapshot, setUp, tearDown, 0, NULL)
 
     return MUNIT_OK;
 }
+
 /* If a candidate server receives a request containing the same term as its
  * own, it it steps down to follower and accept the request . */
 TEST(replication, recvCandidateSameTerm, setUp, tearDown, 0, NULL)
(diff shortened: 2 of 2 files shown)


src/replication.c
```diff
@@ -911,10 +911,16 @@ static void appendFollowerCb(struct raft_io_append *req, int status)
         }
     }
 
-    if (r->state != RAFT_FOLLOWER) {
-        tracef("local server is not follower -> don't send result");
+    /* If our term number has changed since receiving these entries, current_leader
+     * might have changed as well, so don't send a response. */
+    if (r->current_term != args->term) {
```
MathieuBordere, line 916: A node can move to `RAFT_CANDIDATE` without updating its term due to the pre-vote mechanism see
https://github.com/canonical/raft/blob/425a6b57bdea8a600402eac6ee1ab1ea97b36215/src/election.c#L122 

So probably we need something more like `r->current_term != args->term || r->state != RAFT_FOLLOWER`

src/replication.c
```diff
@@ -911,10 +911,16 @@ static void appendFollowerCb(struct raft_io_append *req, int status)
...
+        tracef("new term since receiving entries -> don't respond");
         goto out;
     }
+    /* If our term number hasn't changed, we should still be either FOLLOWER
+     * or UNAVAILABLE -- and the latter is handled by bailing out earlier in
+     * this function. */
+    assert(r->state == RAFT_FOLLOWER);
```
MathieuBordere, line 923: This will break to due remark above.

test/integration/test_replication.c
```diff
@@ -1016,3 +1017,31 @@ TEST(replication, diskWriteFailure, setUp, tearDown, 0, NULL)
...
+    BOOTSTRAP_START_AND_ELECT;
+    CLUSTER_APPLY_ADD_X(0, req, 1, NULL);
+
+    /* Wait for the leader to replicate the entry */
+    CLUSTER_STEP_UNTIL_ELAPSED(500);
+
+    /* Force a new term */
+    CLUSTER_DEPOSE;
+    CLUSTER_ELECT(1);
+
```
MathieuBordere, line 1040: We should probably assert here somewhere that the follower's term has updated?

MathieuBordere (approved):
Looks good, small remark in the test.

PR #377, opened 2023-02-28
Fix LZ4_CFLAGS and raft.pc

### LINKED ISSUE #375: Can't find lz4 and can't disable it
I'm trying to build c-raft. I have lz4 installed and available for pkg-config to find:

```
$ pkg-config --libs --cflags liblz4
-I/projects/spack/opt/spack/linux-ubuntu22.04-sandybridge/gcc-11.3.0/lz4-1.9.4-64syhqtoz2e5finrv5ejiv3mlwu5bu5i/include -L/projects/spack/opt/spack/linux-ubuntu22.04-sandybridge/gcc-11.3.0/lz4-1.9.4-64syhqtoz2e5finrv5ejiv3mlwu5bu5i/lib -llz4
```

When I run c-raft's `configure`, it correctly finds it:

```
...
checking for LZ4... yes
...
```

However when I run `make`, I get the following:

```
  CC       src/libraft_la-byte.lo
  CC       src/libraft_la-client.lo
  CC       src/libraft_la-compress.lo
../src/compress.c:4:10: fatal error: lz4frame.h: No such file or directory
    4 | #include <lz4frame.h>
      |          ^~~~~~~~~~~~
compilation terminated.
make: *** [Makefile:2120: src/libraft_la-compress.lo] Error 1
```

The lz4frame.h is in the include path of lz4.

Additionally, if I configure with `--disable-lz4`, the configure script will still find lz4, enable it, and fail just the same way.

The only way I can build c-raft currently is by (1) making sure that `pkg-config` cannot find lz4 (i.e. removing its path from `PKG_CONFIG_PATH`) and configure with `--disable-lz4`.

There is also an inconsistency in how lz4 and libuv are treated: if libuv is not found, `configure` will simply disable support for it (equivalent to adding `--disable-uv`). But if lz4 is not found, `configure` will fails, unless `--disable-lz4` was explicitly set.

Note: tested with latest release (0.17.1) and master branch.

### LINKED ISSUE #376: Incorrect raft.pc
If c-raft was built with libuv support, the generated raft.pc file should include the following line:

```
Requires: libuv
```

Similarly this line should include `liblz4` if c-raft was built with lz4 support.

cole-miller:
Fixes #375 
Fixes #376

@mdorier, would you mind seeing if this branch fixes the issues you're seeing?

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -78,9 +78,9 @@ test_unit_core_CFLAGS = $(AM_CFLAGS) -Wno-conversion
 test_unit_core_LDADD = libtest.la
 
 if LZ4_AVAILABLE
-test_unit_core_CFLAGS += -DLZ4_AVAILABLE
+test_unit_core_CFLAGS += -DLZ4_AVAILABLE $(LZ4_CFLAGS)
 test_unit_core_LDFLAGS = $(LZ4_LIBS)
-libraft_la_CFLAGS += -DLZ4_AVAILABLE
+libraft_la_CFLAGS += -DLZ4_AVAILABLE $(LZ4_CFLAGS)
 libraft_la_LDFLAGS += $(LZ4_LIBS)
 endif # LZ4_AVAILABLE
 if LZ4_ENABLED
--- a/configure.ac
+++ b/configure.ac
@@ -174,5 +174,10 @@ CC_CHECK_FLAGS_APPEND([AM_LDFLAGS],[LDFLAGS],[ \
 ])
 AC_SUBST(AM_LDLAGS)
 
+REQUIRES=''
+AS_IF(test "x$enable_lz4" != "xno" -a "x$have_lz4" = "xyes", REQUIRES=lz4)
+AS_IF(test "$have_uv" = yes, REQUIRES="$REQUIRES libuv")
+AC_SUBST(REQUIRES)
+
 AC_CONFIG_FILES([raft.pc Makefile])
 AC_OUTPUT
--- a/raft.pc.in
+++ b/raft.pc.in
@@ -6,5 +6,6 @@ includedir=@includedir@
 Name: raft
 Description: C implementation of the Raft Consensus protocol
 Version:  @PACKAGE_VERSION@
+Requires.private: @REQUIRES@
 Libs: -L${libdir} -lraft
 Cflags: -I${includedir}


mdorier:
It doesn't entirely fixes everything:

- [ ]  Without LZ4 installed and without `--disable-lz4`: configure fails (could be improved to silently disable LZ4)
- [X]  Without LZ4 installed and with `--disable-lz4`: configure/make succeed
- [X]  With LZ4 installed and without `--disable-lz4`: configure/make succeed, lz4 detected and used
- [ ]  With LZ4 installed with `--disable-lz4`: configure still finds LZ4 and uses it.

For the last case, it looks like the CFLAGS and LDFLAGS of LZ4 are added, as well as `-DLZ4_AVAILABLE`, but not `-DLZ4_ENABLED`. In the code, `LZ4_AVAILABLE` is used to check whether to use LZ4. Ideally pkg-config should be used to search for LZ4 only if LZ4 was enabled in the first place.

mdorier:
The raft.pc file also still has a problem: the dependency for lz4 should be `liblz4`, not `lz4`.

freeekanayaka:
> @mdorier Thanks for testing.
> 
> > For the last case, it looks like the CFLAGS and LDFLAGS of LZ4 are added, as well as -DLZ4_AVAILABLE, but not -DLZ4_ENABLED. In the code, LZ4_AVAILABLE is used to check whether to use LZ4. Ideally pkg-config should be used to search for LZ4 only if LZ4 was enabled in the first place.
> 
> I think it's intentional that liblz4 is discovered and linked in even if --disable-lz4 is passed. See the comments here:
> 
> https://github.com/canonical/raft/blob/29ab8c5219e41d4cbe4e6b084a9f58837b816c89/configure.ac#L61-L66
> 
> The naming of the flags is a bit misleading.

It sounds like a naming glitch indeed. I think we should ideally have ```--with-lz4```/```--without-lz4``` as a switch that tells whether to use the lz4 **package** (meaning linking against it), and ```--enable-lz4```/```--disable-lz4``` as a switch that tells weather to activate the lz4 **feature** (meaning the "compression on by default" feature). That'd be the conventional naming/semantics with autotools, afaik.

In particular: ```--without-lz4``` means "don't use lz4 at all and don't link against it, regardless if it's available or not", while ```--disable-lz4``` means don't enable lz4 compression by default (but make it optional as long as lz4 is available, either because ```--with-lz4``` was passed or because it was automatically detected). Similarly, ```--with-lz4``` means "do use lz4, and make `./configure` fail if it's not available", and ```--enable-lz4``` means make lz4 compression the default (and likewise fail if lz4 is not available at all). Invalid combinations such as ```--without-lz4 --enable-lz4``` should make `./configure` fail.

Hope my wording wasn't too confusing :)

See https://www.edwardrosten.com/code/autoconf/#nicer-dependencies-with---with-and---enable for a possibly better explanation.

> > The raft.pc file also still has a problem: the dependency for lz4 should be liblz4, not lz4.
> 
> Right you are, I could have sworn I tested `pkg-config --libs lz4` but clearly I didn't.
> 
> > Without LZ4 installed and without --disable-lz4: configure fails (could be improved to silently disable LZ4)
> 
> I'm personally not convinced we should make this change, but I'll let @MathieuBordere weigh in when he returns tomorrow.

I believe the rational of that is that we would like users to use lz4, and avoid them "forgetting" to install lz4. In that sense, it seems good that `./configure` fails if you don't explicitly pass ```--disable-lz4```. Basically we make the ```--with-lz4``` switch implicitly on if it's not passed, which is a pretty common autotools idiom for this kind of things.

mdorier:
Gotcha. My concern is mainly regarding reproducibility and reproducible builds, which is becoming important in research. Just wanted to make sure there is a way to get a deterministic behavior out of the configure script.

freeekanayaka:
> Gotcha. My concern is mainly regarding reproducibility and reproducible builds, which is becoming important in research. Just wanted to make sure there is a way to get a deterministic behavior out of the configure script (by deterministic I mean respecting options independently of what is installed on the system).

Right. Basically you have to decide what your "default" is in case a particular option is not given, and be deterministic about it. Our default is ```--with-lz4``` (or ```--enable-lz4```, as it's called now) and we rightfully fail if we don't find lz4 it and ```--without-lz4``` was not passed.

freeekanayaka:
So, I would update the checklist more or less like this:

- [X ]  Without LZ4 installed and without `--without-lz4`: configure fails (this is what we want)
- [X]  Without LZ4 installed and with `--without-lz4`: configure/make succeed
- [X]  With LZ4 installed and without `--without-lz4`: configure/make succeed, lz4 detected and used
- [ ]  With LZ4 installed with `--without-lz4`: configure/make succeed, lz4 detected but not linked or used at all
- [ ]  With LZ4 installed with `--disable-lz4`: configure/make succeed, links to lz4 but does not enable compression by default (you have to use the API to turn it on).

PR #392, opened 2023-04-03
Simplify queue macros unit tests

freeekanayaka:
Unit tests in `test_queue.c` were using static declaration to workaround a GCC compiler warning about dangling pointers (basically the addresses of the `struct item` objects declared on the stack of the unit test function were stored in the fixture's `queue`, making them dangling when the unit test function returned).

This branch introduces a fixture-level `struct item` array that individual unit tests can safely leverage without needing to explicitly declare it and without ending up with dangling pointers.


freeekanayaka:
> I realize it's a bit ridiculous, but we do need you to sign the CLA using your current email address and add a Signed-off-by line. Otherwise this looks good to me, thanks!

Yeah, I had done that but was using the wrong email :) Pushed another version now, let's see.

PR #398, opened 2023-04-05
membership: Allocate a new raft_io_send to implement raft_transfer

### LINKED ISSUE #396: Potential use-after-free in handling of raft_transfer
When raft_transfer is called, if the UvSend connection pool is all occupied, we end up putting a message in the "pending" queue that holds a pointer to the transfer object:

https://github.com/canonical/raft/blob/c36072c4606aeb2fd53a2f66bf1e3378cebf0e2f/src/membership.c#L217

https://github.com/canonical/raft/blob/c36072c4606aeb2fd53a2f66bf1e3378cebf0e2f/src/uv_send.c#L214

The pending transfer has an associated timeout, and if that expires, we abort the attempt:

https://github.com/canonical/raft/blob/c36072c4606aeb2fd53a2f66bf1e3378cebf0e2f/src/tick.c#L232-L239

And invoke the callback:

https://github.com/canonical/raft/blob/c36072c4606aeb2fd53a2f66bf1e3378cebf0e2f/src/membership.c#L226-L234

But the send request associated with the transfer is still sitting in the pending queue, and so its callback might get invoked a second time, for example when evicting the last message here:

https://github.com/canonical/raft/blob/c36072c4606aeb2fd53a2f66bf1e3378cebf0e2f/src/uv_send.c#L312-L330

This can quickly lead to a use-after-free when the transfer callback owns some memory; I ran into it while debugging https://github.com/canonical/dqlite/pull/480.

cole-miller:
This is @freeekanayaka's suggested patch. Closes #396.

### DIFF
--- a/src/membership.c
+++ b/src/membership.c
@@ -4,6 +4,7 @@
 #include "assert.h"
 #include "configuration.h"
 #include "err.h"
+#include "heap.h"
 #include "log.h"
 #include "progress.h"
 #include "tracing.h"
@@ -194,10 +195,16 @@ void membershipLeadershipTransferInit(struct raft *r,
     r->transfer = req;
 }
 
+static void membershipLeadershipSendCb(struct raft_io_send *send, int status) {
+    (void)status;
+    RaftHeapFree(send);
+}
+
 int membershipLeadershipTransferStart(struct raft *r)
 {
     const struct raft_server *server;
     struct raft_message message;
+    struct raft_io_send *send;
     int rv;
     assert(r->transfer->send.data == NULL);
     server = configurationGet(&r->configuration, r->transfer->id);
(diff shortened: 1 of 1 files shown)


MathieuBordere:
> Working on a regression test that will trigger the UAF in the old code.

A test would be great indeed, cause I'm not convinced this will fix it, it's possible I will have an aha-moment after typing this comment (partially stolen from Cole), so please bear with me :-)

When `raft_transfer` is called, if the `UvSend` connection pool is all occupied, we end up putting a message in the "pending" queue that holds a pointer to the transfer object:

https://github.com/canonical/raft/blob/c6cfe587fc5fa26e78e76c8c32f33e510d01cfda/src/membership.c#L245

https://github.com/canonical/raft/blob/c36072c4606aeb2fd53a2f66bf1e3378cebf0e2f/src/uv_send.c#L214

The pending transfer has an associated timeout, and if that expires, we abort the attempt:

https://github.com/canonical/raft/blob/c36072c4606aeb2fd53a2f66bf1e3378cebf0e2f/src/tick.c#L232-L239

And invoke the callback:
https://github.com/canonical/raft/blob/c6cfe587fc5fa26e78e76c8c32f33e510d01cfda/src/membership.c#L255-L263

And free the send:
https://github.com/canonical/raft/blob/c6cfe587fc5fa26e78e76c8c32f33e510d01cfda/src/membership.c#L198-L201

But the allocated, and now freed send request associated with the transfer is still sitting in the pending queue and we access it's cb field, leading to accessing free'd memory.

https://github.com/canonical/raft/blob/c36072c4606aeb2fd53a2f66bf1e3378cebf0e2f/src/uv_send.c#L320-L326

It's plausible I'm missing something here.

MathieuBordere (approved):

PR #399, opened 2023-04-10
clang-format of raft.h

freeekanayaka:
This branch adjusts a few comments and docstrings in `raft.h` to fit the 80 characters line limit, and runs `clang-format` against the whole file.

We have `.clang-format` file in the top-level directory of the source tree. It's basically the equivalent of `gofmt` in Golang, and helps keep the code layout consistent. We might tweak the configuration options in `.clang-format` if there are styles we want to change.

We might also consider adding a Github workflow lint action like https://github.com/marketplace/actions/clang-format-lint, or something along those lines.

### DIFF
--- a/include/raft.h
+++ b/include/raft.h
@@ -352,8 +352,8 @@ struct raft_message
  * Hold the details of a snapshot.
  * The user-provided raft_buffer structs should provide the user with enough
  * flexibility to adapt/evolve snapshot formats.
- * If this struct would NEED to be adapted in the future, raft can always move to
- * a new struct with a new name and a new raft_io version.
+ * If this struct would NEED to be adapted in the future, raft can always move
+ * to a new struct with a new name and a new raft_io version.
  */
 struct raft_snapshot
 {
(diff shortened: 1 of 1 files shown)


freeekanayaka:
> I think if we do a clang-format of the whole tree, we should standardize on one convention for function and type names while we're at it and fix the whole codebase to conform to that style. Happy to take that on if I've got your and Mathieu's buy-in.

I previewed the diff of a clang-format of the whole tree, it seems 99% ok, but I noticed a few comments or docstring lines exceeding the 80 chars limit that were not reformatted perfectly (because they had somehow custom layouts like bullet points and such).

It's kinda minor, so I'd be okay committing such a diff and then fixing those comments separately, either in a separate follow-up commit in the same PR, or even opportunistically later on. The important thing would be to keep commits from automated clang-format run separates from commits with manual changes, so we don't miss anything changed accidentally by hand.

Regarding function and type names, I feel it should be done incrementally and not in bulk, so we have a chance to review it a bit more carefully.

PR #401, opened 2023-04-12
raft.h: Introduce raft_tracer_v2 for more flexible tracing

cole-miller:
Here is a stab at introducing a more flexible tracing scheme for raft, that should hopefully support our plans to revamp and integrate tracing across raft and dqlite. It is backwards-compatible.

### DIFF
--- a/include/raft.h
+++ b/include/raft.h
@@ -435,6 +435,9 @@ struct raft_io_async_work
     raft_io_async_work_cb cb;   /* Request callback */
 };
 
+#define RAFT_TRACER_EMIT_V2 \
+    ((void (*)(struct raft_tracer *, const char *, int, const char *))1)
+
 /**
  * Customizable tracer, for debugging purposes.
  */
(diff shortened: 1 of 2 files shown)


MathieuBordere:
> > `Error: test/lib/heap.c:138: memory leak: -1 outstanding allocations`
> 
> That's weird, maybe something to do with #397?



> > `Error: test/lib/heap.c:138: memory leak: -1 outstanding allocations`
> 
> That's weird, maybe something to do with #397?

Probably, the leak detection is pretty basic. The original code checked if there are the same number of `alloc` and `free` calls, #397 violates that assumption. We could revert #397 and then check that we don't do 0-size allocations.

MathieuBordere:
Looks okay, before merging this it probably makes senseif we directly start handling the new tracer version in the tracing code?

freeekanayaka:
I'd like to have a broader picture of the use cases and requirements, so we make sure we are not missing something. For example, the point @cole-miller makes about metrics seems important: if we decide that tracing and metrics should share tooling, then it'd be good to have an idea of what will be needed and if any accommodation is due (my gut feeling is that metrics is something separate that won't share tooling, but I didn't quite reason on it).

freeekanayaka:
> > > `Error: test/lib/heap.c:138: memory leak: -1 outstanding allocations`
> > 
> > 
> > That's weird, maybe something to do with #397?
> 
> > > `Error: test/lib/heap.c:138: memory leak: -1 outstanding allocations`
> > 
> > 
> > That's weird, maybe something to do with #397?
> 
> Probably, the leak detection is pretty basic. The original code checked if there are the same number of `alloc` and `free` calls, #397 violates that assumption. We could revert #397 and then check that we don't do 0-size allocations.

Yes, that leak detection is mainly a sanity check for what I'd say should be the good practice of performing only greater-than-zero allocations and releasing the one and only times. The 0-size allocations might save a line or two of code in some cases, but I'm not sure it's worth, since it's quite handy to know that allocations and deallocations are balanced, as a first data point before further debugging a memory leak. To further debug a memory leak triggered during the test suite then the `--enable-sanitize` configure option is what should be used at the moment.

PR #404, opened 2023-04-14
compress.c: Allow to not use LZ4 even if available

prrlvr:
Hello,

I encountered this issue yesterday : I would like to use the lib without linking to LZ4, even if it is present on my system.
Hopefully it doesn't change much.

As `LZ4_ENABLED` is on by default if `LZ4_AVAILABLE` is defined, it does not change the previous behavior.

### DIFF
--- a/src/compress.c
+++ b/src/compress.c
@@ -19,7 +19,7 @@ int Compress(struct raft_buffer bufs[],
              struct raft_buffer *compressed,
              char *errmsg)
 {
-#ifndef LZ4_AVAILABLE
+#if !defined(LZ4_AVAILABLE) || !defined(LZ4_ENABLED)
     (void)bufs;
     (void)n_bufs;
     (void)compressed;
@@ -169,7 +169,7 @@ int Decompress(struct raft_buffer buf,
                struct raft_buffer *decompressed,
                char *errmsg)
 {
-#ifndef LZ4_AVAILABLE
+#if !defined(LZ4_AVAILABLE) || !defined(LZ4_ENABLED)
     (void)buf;
     (void)decompressed;
     ErrMsgPrintf(errmsg, "LZ4 not available");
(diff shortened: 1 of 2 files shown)


MathieuBordere:
Passing `--disable-lz4` will stop raft from compressing snapshots, however when it tries to read a compressed snapshot from disk, it will still be able to decompress it. Does that suit your needs?

prrlvr:
The idea was more to be able to use the lib without linking to `liblz4`. That sounds odd if `IsCompressed` returns true even when the decompress code is disabled, but that should not happen.

MathieuBordere:
Will take a look this week @prrlvr , I was away last week.

freeekanayaka:
> @prrlvr what about a new flag for the configure script that causes LZ4_AVAILABLE to be unconditionally unset, i.e. without consulting pkg-config? That would preserve the desirable feature that Mathieu mentioned, in the default configuration.

+1

The convention with autotools would be that `--with-<dependency>` and `--wihtout-<dependency>` control whether to link to `<dependency>` (e.g. `--with-lz4` and `--without-lz4`), while `--enable-<feature>` and `--disable-<feature>` control weather to activate a feature (e.g. `--enable-compression` and `--disable-compression`). In this case since compression is desirable, `--with-lz4` and `--enable-compression` would both default to "yes" in case `libz4` is found, and gracefully fallback to no only if not found, without failing. However if `--with-lz4` is explicitly passed, then `./configure` would fail if `libz4` is not found, while an explicit `--without-lz4` would prevent from linking to `libz4` regardless of whether `libz4` is found or not. The `--enable-compression` option would be a bit orthogonal to that: assuming that `libz4` is linked, it would basically control wether compression is on by default or not (in both cases you could use `raft_uv_set_snapshot_compression` to change the default).

See also https://autotools.info/autoconf/arguments.html.

PR #409, opened 2023-04-18
Slight simplification of configuration rollback logic

freeekanayaka:
This change removes a bit of code duplication in the logic that handles configuration rollbacks when the last committed configuration is not available anymore in the log and must be retrieved from the last snapshot.

It basically makes the code reflect a bit more [this comment](https://github.com/canonical/raft/issues/250#issuecomment-1329249557) from @cole-miller that was made when the fix for configuration rollbacks was implemented.

Please see the commit messages of the individual commits in the PR for more details.

The diff removes more code than it adds, yet test coverage should slightly increase due to reduced code duplication.


freeekanayaka:
> > Thanks, always happy to see some redundancy removed! To be clear, r->configuration_index still refers to the index of the last committed configuration whether or not it's part of a snapshot, right?
> 
> Exactly, see the associated [comment](https://github.com/canonical/raft/blob/9fc0c0b88c9fa72f61941f760be76da2ad216701/include/raft.h#L633) in the public header, I had improved them [here](https://github.com/canonical/raft/commit/dc6dd609e5db6767f67ba6a77a8392e43c17b84b).

We might even consider renaming `configuration_index` to `configuration_committed_index`, if that helps.

freeekanayaka:
> > > > Thanks, always happy to see some redundancy removed! To be clear, r->configuration_index still refers to the index of the last committed configuration whether or not it's part of a snapshot, right?
> > > 
> > > 
> > > Exactly, see the associated [comment](https://github.com/canonical/raft/blob/9fc0c0b88c9fa72f61941f760be76da2ad216701/include/raft.h#L633) in the public header, I had improved them [here](https://github.com/canonical/raft/commit/dc6dd609e5db6767f67ba6a77a8392e43c17b84b).
> > 
> > 
> > We might even consider renaming `configuration_index` to `configuration_committed_index`, if that helps.
> 
> I'd support that!

Ok then, I'll make a separate PR for that.

PR #412, opened 2023-05-04
raft.h: Allow overriding RAFT_API

cole-miller:
See also libuv:

https://github.com/libuv/libuv/blob/3f331e97da5b1cc2e24765dcb6d26c9dd313418f/include/uv.h#L34

### DIFF
--- a/include/raft.h
+++ b/include/raft.h
@@ -7,7 +7,9 @@
 #include <stdint.h>
 #include <stdio.h>
 
+#ifndef RAFT_API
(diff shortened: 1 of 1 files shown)


MathieuBordere:
out of curiosity, what do you plan to do with that?

freeekanayaka:
This is ok, however as a general rule I'd wait for real-world use cases to arise before introducing this kind of changes. Expecially in a world where apps are more and more confined and containerized, this is probably not an issue in production.

PR #421, opened 2023-05-31
Respond if term changed when installing snapshot

### LINKED ISSUE #355: Assertion `r->last_applied <= r->commit_index' failed.
Observed during Jepsen runs:

https://github.com/canonical/jepsen.dqlite/actions/runs/3873600397/jobs/6603811458

https://github.com/canonical/jepsen.dqlite/actions/runs/3869648631/jobs/6595906665

https://github.com/canonical/jepsen.dqlite/actions/runs/3860783772/jobs/6581330548

Looks like it should have been introduced by recently merged PR's as it's occurring frequently and has only started recently. Based on the first run in which it occurs (the last one in the list aboive), it's possible it's introduced by #351, #352 or #353.

freeekanayaka:
This change simplifies slightly the logic for handling completions of `raft_io->snapshot_put()` requests on followers. It stops tracking the term that was in place when the request started, and instead just looks at the current situation, since it's not only harmless to send and AppendEntries RPC result to a different leader, but it might actually be useful since a new leader might get to now about the follower's state faster.

I don't expect this PR alone to fix #355, but it's an isolated change that I believe makes sense on its own. If merged, we can wait a bit to see how Jepsen behaves (I'd expect no improvements but also no regressions).

Next step would be to do a similar change for the equivalent logic that we have after completing a `raft_io->append()` request. Then I'd like also to allow converting to candidate state while installing a snapshot, for consistency with what we do when appending entries and also because it's probably going to simplify logic and avoid unnecessary delays in elections. We can push these changes slowly to observe jepsen's reactions.

### DIFF
--- a/src/replication.c
+++ b/src/replication.c
@@ -839,6 +839,23 @@ static void sendAppendEntriesResult(
     int rv;
 
     assert(r->state == RAFT_FOLLOWER);
+
+    /* There are two cases in which a follower can have no leader:
+     *
+     * - If it never had a leader before (e.g. it just started)
+     *
+     * - If has become a follower after stepping down from leader because
+     *   it could not contact a majority of servers.
+     *
+     * In the first case we can't reach this function because no entries have
+     * been received.
+     *
+     * In the second case we don't call this function because we don't send
+     * AppendEntries result when completing a raft_io->append() request
+     * initiated by a leader.
+     */
+    assert(r->follower_state.current_leader.address != NULL);
+
     message.type = RAFT_IO_APPEND_ENTRIES_RESULT;
     message.server_id = r->follower_state.current_leader.id;
     message.server_address = r->follower_state.current_leader.address;
--- a/src/replication.c
+++ b/src/replication.c
@@ -1233,7 +1233,6 @@ struct recvInstallSnapshot
 {
     struct raft *raft;
     struct raft_snapshot snapshot;
-    raft_term term; /* Used to check for state transitions. */
 };
 
 static void installSnapshotCb(struct raft_io_snapshot_put *req, int status)
@@ -1242,7 +1241,6 @@ static void installSnapshotCb(struct raft_io_snapshot_put *req, int status)
     struct raft *r = request->raft;
     struct raft_snapshot *snapshot = &request->snapshot;
     struct raft_append_entries_result result;
-    bool should_respond = true;
     int rv;
 
     /* We avoid converting to candidate state while installing a snapshot. */
@@ -1258,18 +1256,7 @@ static void installSnapshotCb(struct raft_io_snapshot_put *req, int status)
     /* If we are shutting down, let's discard the result. */
     if (r->state == RAFT_UNAVAILABLE) {
         tracef("shutting down -> discard result of snapshot installation");
-        should_respond = false;
         goto discard;
-    }
-    /* If the request is from a previous term, it means that someone else became
-     * a candidate while we were installing the snapshot. In that case, we want
-     * to install the snapshot anyway, but our "current leader" may no longer be
-     * the same as the server that sent the install request, so we shouldn't
-     * send a response to that server. */
-    if (request->term != r->current_term) {
-        tracef(
-            "new term since receiving snapshot -> install but don't respond");
-        should_respond = false;
     }
 
     if (status != 0) {
@@ -1304,7 +1291,7 @@ discard:
     raft_configuration_close(&snapshot->configuration);
 
 respond:
-    if (should_respond) {
+    if (r->state == RAFT_FOLLOWER) {
         result.last_log_index = r->last_stored;
         sendAppendEntriesResult(r, &result);
     }
@@ -1364,7 +1351,6 @@ int replicationInstallSnapshot(struct raft *r,
         goto err;
     }
     request->raft = r;
-    request->term = r->current_term;
 
     snapshot = &request->snapshot;
     snapshot->term = args->last_term;
--- a/test/integration/test_replication.c
+++ b/test/integration/test_replication.c
@@ -1210,3 +1210,42 @@ TEST(replication, failPersistBarrier, setUp, tearDown, 0, NULL)
 
     return MUNIT_OK;
 }
+
+/* A follower finds that is has no leader anymore after it completes persisting
+ * entries. No AppendEntries RPC result is sent in that case. */
+TEST(replication, noLeaderAfterAppending, setUp, tearDown, 0, NULL)
+{
+    struct fixture *f = data;
+    struct raft_apply *req = munit_malloc(sizeof(*req));
+
+    /* Make sure that persisting entries on server 0 will take a long time. */
+    CLUSTER_SET_DISK_LATENCY(0, 1100);
+
+    /* Server 0 becomes the leader. */
+    BOOTSTRAP_START_AND_ELECT;
+    munit_assert_int(CLUSTER_STATE(0), ==, RAFT_LEADER);
+
+    /* Submit a new entry, which server 0 will start to persist. */
+    CLUSTER_APPLY_ADD_X(0, req, 1, NULL);
+
+    /* Disconnect server 0, so it will step down and become follower. */
+    CLUSTER_DISCONNECT(0, 1);
+    CLUSTER_DISCONNECT(1, 0);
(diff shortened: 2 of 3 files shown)


MathieuBordere:
There's this feature in the jepsen repo that allows to run the tests against a custom raft branch, see https://github.com/canonical/jepsen.dqlite/pull/82

freeekanayaka:
> There's this feature in the jepsen repo that allows to run the tests against a custom raft branch, see [canonical/jepsen.dqlite#82](https://github.com/canonical/jepsen.dqlite/pull/82)

Nice.

I've browsed to https://github.com/canonical/jepsen.dqlite/actions/workflows/test-sanitize.yml?query=, but I can't find the `Run workflow` button. Here's what I see:

Screenshot from 2023-05-31 13-53-45

A single or few runs would be nice, but I think we'll have to wait a day or two to see some statistical data.

MathieuBordere:
> It's okay to send an AppendEntries result to a leader that was not the same of
the one that originally sent the snapshot, or that has bumped its term.

Conceptually I don't think I can agree with this, sending a response to a server that didn't send a request.

freeekanayaka:
> > It's okay to send an AppendEntries result to a leader that was not the same of
> > the one that originally sent the snapshot, or that has bumped its term.
> 
> Conceptually I don't think I can agree with this, sending a response to a server that didn't send a request.

I believe we shouldn't regard them as plainly "regular" requests in the traditional sense. These "requests" are actually "messages" or "events", for example they could be duplicated and they are broadly idempotent. The "response" here is really just "let the current leader know about my state". It's basically message-driven state synchronization.

In general, trying to minimize the state we track leads to simpler logic to reason around, since in most cases looking at the situation as it is now is less complicated than looking at the situation as it is now combined with what the situation was at some point the past. I believe we have a few cases where things can be improved in that regard.

freeekanayaka:
> I've browsed to https://github.com/canonical/jepsen.dqlite/actions/workflows/test-sanitize.yml?query=, but I can't find the `Run workflow` button. Here's what I see:

I figured this out, I was looking at "Actions -> Dqlite Jepsen tests", but it's actually "Actions -> Callabale Dqlite Jepsen tests".

freeekanayaka:
> I don't have a strong opinion about sending responses to the "wrong" server.

I'll add a couple of additional observations here to possibly address the conceptual concern:

- The semantics of an AppendRequest message is basically: "follower, please append these entries and when done notify your progress to the leader". If the leader changes, the new leader is not the "wrong" server, it's actually precisely the server that the follower should notify, so it's the "right" server. Note that since this is a message/event oriented synchronization protocol there's no mandatory coupling or pairing between messages (unlike say, HTTP).

- As we know, the message/event oriented design supports unreliable networks where messages can be delayed, duplicated or dropped. Imagine leader L1 sending entry E1 to follower F, then leader L1 steps down and leader L2 gets elected. L2 also sends the exact E1 entry to follower F. Then L2 receives a notification message from F that it has appended E1. This notification message could have been triggered by the original message that L1 sent, and that F sent to L2 instead of L1 because it completed persisting the entry after learning about L2, or it could have been triggered by the new message that L2 sent to F: if one of the two notification messages was dropped by the network, there would be no way for L2 to distinguish between the "right" one and the "wrong" one, and rightfully so, because actually there is no right or wrong, there is just a follower fulfilling its expected behavior.

> Just need to make sure that #339 is not reintroduced in some form.

This PR basically applies the same `if (r->state == RAFT_FOLLOWER)` check as #342, which is enough for #339, because we now send a message only if we are followers and hence `r->follower_state.current_leader` is not corrupted. There are also additional assertions now in [this commit](https://github.com/canonical/raft/pull/421/commits/6b3c53508cdddaf5765a6e74762f8ad632c9d9bc) to signal any future violation that would lead to a regression with respect to #339.

MathieuBordere:
> > > It's okay to send an AppendEntries result to a leader that was not the same of
> > > the one that originally sent the snapshot, or that has bumped its term.
> > 
> > 
> > Conceptually I don't think I can agree with this, sending a response to a server that didn't send a request.
> 
> I believe we shouldn't regard them as plainly "regular" requests in the traditional sense. These "requests" are actually "messages" or "events", for example they could be duplicated (a receiver can get the same message multiple times, because the network duplicates it) and they are broadly idempotent. The "response" here is really just a "let the current leader know about my state" message. It's basically message-driven state synchronization. Thinking about messages probably solves the conceptual side.
> 
> In general, trying to minimize the state we track leads to simpler logic to reason around, since in most cases looking at the situation as it is now is less complicated than looking at the situation as it is now combined with what the situation was at some point the past. I believe we have a few cases where things can be improved in that regard.

I'm just a bit worried that by sending an AppendEntriesRPC to a node, the leader sets some state and that the processing of the AppendEntriesResultRPC assumes that particular state. So when we send a Result to a different leader, that leader might not expect that particular answer and things go wrong. I would need to look into it to ease my mind a bit.

freeekanayaka:
> I'm just a bit worried that by sending an AppendEntriesRPC to a node, the leader sets some state and that the processing of the AppendEntriesResultRPC assumes that particular state. So when we send a Result to a different leader, that leader might not expect that particular answer and things go wrong. I would need to look into it to ease my mind a bit.

That's precisely the type of state-related complication this PR tries to reduce, and I think we have a few cases like this that we can incrementally simplify.

The less assumptions/state we have and track, especially about the past, the less complex the resulting logic will be, and reasoning about the associated mechanics becomes easier.

Note that a node must be prepared to receive any message from any node at any time, because the network is unreliable (a TCP transport is an implementation detail, it could be, say, UDP) and can duplicate, drop or delay messages arbitrarily.

Essentially `struct raft` is a deterministic state machine that takes in an event (e.g. a message), looks at its **current** state, and outputs some other messages to be sent, or data to be written to disk.

In this particular case, the leader might indeed set some internal state associated to a AppendEntriesRPC message, but it needs to be prepared to receive any AppendEntriesResultRPC from anyone at any time, because for example the AppendEntriesResultRPC  it receives might be a delayed message an in the meantime the leader had stepped down, reset that internal state associated with that initial AppendEntriesRPC and then become leader again with a brand new state. Just an example to illustrate that no assumptions can be made that strongly couple messages, and that, being this a deterministic event-driver state machine, reasoning on the current state only is generally both simpler and more robust.

MathieuBordere:
> > I'm just a bit worried that by sending an AppendEntriesRPC to a node, the leader sets some state and that the processing of the AppendEntriesResultRPC assumes that particular state. So when we send a Result to a different leader, that leader might not expect that particular answer and things go wrong. I would need to look into it to ease my mind a bit.
> 
> That's precisely the type of state-related complication this PR tries to reduce, and I think we have a few cases like this that we can incrementally simplify.
> 
> The less assumptions/state we have and track, especially about the past, the less complex the resulting logic will be, and reasoning about the associated mechanics becomes easier.
> 
> Note that a node must be prepared to receive any message from any node at any time, because the network is unreliable (a TCP transport is an implementation detail, it could be, say, UDP) and can duplicate, drop or delay messages arbitrarily.
> 
> Essentially `struct raft` is a deterministic state machine that takes in an event (e.g. a message), looks at its **current** state, and outputs some other messages to be sent, or data to be written to disk.
> 
> In this particular case, the leader might indeed set some internal state associated to an AppendEntriesRPC message, but it needs to be prepared to receive any AppendEntriesResultRPC message from anyone at any time, because for example the AppendEntriesResultRPC it receives might be a delayed message an in the meantime the leader had stepped down, reset that internal state associated with that initial AppendEntriesRPC and then become leader again with a brand new state, which is equivalent to the situation you describe, if I understand correctly. This is just an example to illustrate that no assumptions can be made that strongly couple messages, and that, being this a deterministic event-driver state machine, reasoning on the current state only is generally both simpler and more robust, while trying to prevent this kind situation from happening at all by using state to introduce constraints is going to be more fragile and complicated to reason around.

This is where we get into a bit of trouble with our implementation I think, original raft assumes an `RPC` that can succeed or fail, we don't really have that luxury.

freeekanayaka:
> This is where we get into a bit of trouble with our implementation I think, original raft assumes an `RPC` that can succeed or fail, we don't really have that luxury.

I'm not sure to understand exactly what you mean, given that you are sending a message to a remote server, you can't know if it succeeded or it failed (if you don't hear a reply anything could have happened). Or perhaps you mean something else.

PR #427, opened 2023-06-05
Flags

MathieuBordere:
This is preparation for introducing a new message type, `RaftInstallSnapshotChunk` (naming to be decided) for sending a snapshot in parts to a node.

The problem is that the leader has to know beforehand if it can send a snapshot chunk to a node, because a legacy node expects a complete snapshot in a message. In order to do that, the leader has to know the capabilities of the node. I propose to add a `flags` field to an `AppendEntriesResultRPC`. The leader can then decide based on the nodes capabilities set in the `flags` to send it a complete snapshot or only a part of a snapshot.


src/progress.h
```diff
@@ -23,7 +23,8 @@ struct raft_progress
     raft_index snapshot_index;    /* Last index of most recent snapshot sent. */
     raft_time last_send;          /* Timestamp of last AppendEntries RPC. */
     raft_time snapshot_last_send; /* Timestamp of last InstallSnaphot RPC. */
-    bool recent_recv; /* A msg was received within election timeout. */
+    bool recent_recv;       /* A msg was received within election timeout. */
+    uint64_t feature_flags; /* What the server is capable of. */
```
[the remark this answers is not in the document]
MathieuBordere: yeah it should be

test/unit/test_flags.c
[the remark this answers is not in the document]
MathieuBordere: Yeah, it's more of a sanity check to check for obvious bugs.

src/flags.h
```diff
@@ -0,0 +1,21 @@
+#ifndef FLAGS_H_
+#define FLAGS_H_
+
+#include "../include/raft.h"
+
+#define RAFT_FLAGS_FLAGS (1 << 0)  // This node has the 'Flags' feature.
```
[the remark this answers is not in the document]
MathieuBordere: yeah I tried to follow convention a bit, and ended up with a weird name. Any other flag would be `RAFT_FLAGS_SOMEFEATURE`. If you really hate it I can change it :-)

include/raft.h
```diff
@@ -818,8 +819,9 @@ struct raft
     unsigned max_catch_up_rounds;
     unsigned max_catch_up_round_duration;
 
+    raft_flags feature_flags;
```
freeekanayaka, line 822: Do we need to store this value at all in `struct raft`? I'd expect this to be a compile-time constant, in other word `struct raft` will always set `raft_append_entries_result->flags` to whatever the capabilities were at the time the code was compiled. I wouldn't expect `raft->feature_flags` to store a different value than that, unless I'm missing some detail of the mechanics.

include/raft.h
```diff
@@ -79,6 +79,13 @@ typedef unsigned long long raft_index;
  */
 typedef unsigned long long raft_time;
 
+/**
+ * Hold the features a raft node is capable of.
+ * This typedef must always be 8 bytes wide in order not to mess with ABI
+ * compatibility between raft versions.
```
freeekanayaka, line 85: It's a tad unfortunate that the type we use is dictated not strictly by its intent/goal, but by an implementation detail about padding/ABI compatibility. I have an idea about this, perhaps I can submit a follow-up PR to see what people think. Not blocking, just wanted to point that out.

src/flags.h
```diff
@@ -0,0 +1,22 @@
+#ifndef FLAGS_H_
+#define FLAGS_H_
+
+#include "../include/raft.h"
+
+/* This node has the 'Flags' feature. */
+#define RAFT_FLAGS_FLAGS ((raft_flags)1 << 0)
```
freeekanayaka, line 7: Is it necessary to know if the other node has the flags feature?

As far as I understand, if a node running this new code receives an AppendEntries result from a legacy node which is not running this new code, then the `raft_append_entries_result->flags` field will set be `0`, and the node running this new code will avoid sending to the legacy node anything that the legacy node will not understand (e.g. InstallSnapshot chunks). On the other hand, if the nodes that sends the AppendEntries result is also running this newer code it will set `raft_append_entries_result->flags` either to `0` or to some value different than `0`. If it sets it to `0` it means that it doesn't support any new feature and it can be treated exactly like the legacy node that didn't send the flags field at all. If it did set some flags, then the receiving node will take advantage of the flags that it knows how to deal with (e.g. InstallSnapshot chunks).

If for some reason we need also to differentiate between the case in which `raft_append_entries_result->flags` is `0` because the node didn't send it at all, and the case where the node set it to `0`, then `raft_append_entries_result->version` can be used (it will be `0` in the former case and `1` in the latter).

Assuming what I'm saying matches the intended mechanics, I believe that `RAFT_FLAGS_FLAGS` could be dropped, and `RAFT_DEFAULT_FEATURE_FLAGS` could be set to 0 (and changed later on when InstallSnapshot chunks are introduced).

freeekanayaka (commented):
Thanks for breaking up the bigger feature into smaller incremental PRs. Looks good to me overall, left a few comments/questions.

freeekanayaka:
> This is preparation for introducing a new message type, `RaftInstallSnapshotChunk` (naming to be decided) for sending a snapshot in parts to a node.

Unrelated to this PR (which is going to be needed no matter what), but do we need a new message type for chunked snapshots?

I presume that `RaftInstallSnapshotChunk`is going to have exactly the same fields as the current `RaftInstallSnapshot` (aka `struct raft_install_snapshot`) plus the additional `offset` and `done` fields, as per figure 5.3 of the Raft dissertation.

If that's the case, couldn't we just have a single `RaftInstallSnapshot` message type and bump its version to `1`? Basically the same that this PR is doing for `RaftAppendEntriesResult`.

If a legacy node sends a `RaftInstallSnapshot` at version 0 to a newer node that also understands version 1, the newer node can just set `offset` to `0` and `done` to `true`, and its new code for handling chunked snapshots should work just fine also for handling this non-chunked snapshot, which will be treated as a chuncked snapshot with just one chunk (a case which the new node should be able to handle anyways). Probably the only difference will be that when the "chunk" has been persisted by the newer node, and the newer node needs to respond to the legacy node it will respond to it with an `AppendEntriesResult`, instead of a `RaftInstallSnapshotResult` (which is going to be needed for the chunked case).

Conversely if a newer node sends a `RaftInstallSnapshot` to an older node, it can simply use the same "send the snapshot in chunks" code paths and logic that it has and would use otherwise, with the only tweak of setting the chunk size to the size of the snapshot (i.e. the number of chunks will be 1). It can also safely send the message as version 1, since the legacy node will "downgrade" it to version 0, ignoring the missing fields.

In this way we should minimize special casing, and can have a single code path that handles both cases (chuncked and not-chunked `RaftInstallSnapshot`), and in the future we should also be able to drop support for the non-chunked version `0` of `RaftInstallSnapshot` (i.e. drop the code that conditionally replies `AppendEntriesResult` or `RaftInstallSnapshotResult` and the code that sends just 1 chunk instead of N), because enough time will have passed that we can safely assume that all nodes understand version `1`.

freeekanayaka:
I'm wondering if we should include the `flags` field in all messages, not just `AppendEntriesResult`, for future proofing. For example, a follower will receive an `AppendEntries` message that will have contain the leader's feature flags and perhaps modify its behavior accordingly. Same for `RequestVote`/`RequestVoteResult`.

I'm just thinking loud here, not sure.

src/flags.h
MathieuBordere: Yeah, that's true, it was more of a convenience to be able to test it easier in the integration test, otherwise I had to get into more hoops to know if the other node had sent some flags because a 0 value could mean no flags or flags not supported.

MathieuBordere:
> > This is preparation for introducing a new message type, `RaftInstallSnapshotChunk` (naming to be decided) for sending a snapshot in parts to a node.
> 
> Unrelated to this PR (which is going to be needed no matter what), but do we need a new message type for chunked snapshots?
> 
> I presume that `RaftInstallSnapshotChunk`is going to have exactly the same fields as the current `RaftInstallSnapshot` (aka `struct raft_install_snapshot`) plus the additional `offset` and `done` fields, as per figure 5.3 of the Raft dissertation.
> 
> If that's the case, couldn't we just have a single `RaftInstallSnapshot` message type and bump its version to `1`? Basically the same that this PR is doing for `RaftAppendEntriesResult`.
> 
> If a legacy node sends a `RaftInstallSnapshot` at version 0 to a newer node that also understands version 1, the newer node can just set `offset` to `0` and `done` to `true`, and its new code for handling chunked snapshots should work just fine also for handling this non-chunked snapshot, which will be treated as a chuncked snapshot with just one chunk (a case which the new node should be able to handle anyways). Probably the only difference will be that when the "chunk" has been persisted by the newer node, and the newer node needs to respond to the legacy node it will respond to it with an `AppendEntriesResult`, instead of a `RaftInstallSnapshotResult` (which is going to be needed for the chunked case).
> 
> Conversely if a newer node sends a `RaftInstallSnapshot` to an older node, it can simply use the same "send the snapshot in chunks" code paths and logic that it has and would use otherwise, with the only tweak of setting the chunk size to the size of the snapshot (i.e. the number of chunks will be 1). It can also safely send the message as version 1, since the legacy node will "downgrade" it to version 0, ignoring the missing fields.
> 
> In this way we should minimize special casing, and can have a single code path that handles both cases (chuncked and not-chunked `RaftInstallSnapshot`), and in the future we should also be able to drop support for the non-chunked version `0` of `RaftInstallSnapshot` (i.e. drop the code that conditionally replies `AppendEntriesResult` or `RaftInstallSnapshotResult` and the code that sends just 1 chunk instead of N), because enough time will have passed that we can safely assume that all nodes understand version `1`.

Yes I thought about it too recently. At first I thought the `RaftInstallSnapshotRPC` couldn't be versioned because it had dynamic size, but it can be done. Will chew on it for a bit before I open the follow-up PRs.

freeekanayaka:
> > > This is preparation for introducing a new message type, `RaftInstallSnapshotChunk` (naming to be decided) for sending a snapshot in parts to a node.
> > 
> > 
> > Unrelated to this PR (which is going to be needed no matter what), but do we need a new message type for chunked snapshots?
> > I presume that `RaftInstallSnapshotChunk`is going to have exactly the same fields as the current `RaftInstallSnapshot` (aka `struct raft_install_snapshot`) plus the additional `offset` and `done` fields, as per figure 5.3 of the Raft dissertation.
> > If that's the case, couldn't we just have a single `RaftInstallSnapshot` message type and bump its version to `1`? Basically the same that this PR is doing for `RaftAppendEntriesResult`.
> > If a legacy node sends a `RaftInstallSnapshot` at version 0 to a newer node that also understands version 1, the newer node can just set `offset` to `0` and `done` to `true`, and its new code for handling chunked snapshots should work just fine also for handling this non-chunked snapshot, which will be treated as a chuncked snapshot with just one chunk (a case which the new node should be able to handle anyways). Probably the only difference will be that when the "chunk" has been persisted by the newer node, and the newer node needs to respond to the legacy node it will respond to it with an `AppendEntriesResult`, instead of a `RaftInstallSnapshotResult` (which is going to be needed for the chunked case).
> > Conversely if a newer node sends a `RaftInstallSnapshot` to an older node, it can simply use the same "send the snapshot in chunks" code paths and logic that it has and would use otherwise, with the only tweak of setting the chunk size to the size of the snapshot (i.e. the number of chunks will be 1). It can also safely send the message as version 1, since the legacy node will "downgrade" it to version 0, ignoring the missing fields.
> > In this way we should minimize special casing, and can have a single code path that handles both cases (chuncked and not-chunked `RaftInstallSnapshot`), and in the future we should also be able to drop support for the non-chunked version `0` of `RaftInstallSnapshot` (i.e. drop the code that conditionally replies `AppendEntriesResult` or `RaftInstallSnapshotResult` and the code that sends just 1 chunk instead of N), because enough time will have passed that we can safely assume that all nodes understand version `1`.
> 
> Yes I thought about it too recently. At first I thought the `RaftInstallSnapshotRPC` couldn't be versioned because it had dynamic size, but it can be done. Will chew on it for a bit before I open the follow-up PRs.

Yeah the dynamic part is separated from the static part, the static size comes first and it's fully read before the dynamic part. So `InstallSnapshot` is extensible/versionable in the same way as other messages.

src/flags.h
freeekanayaka: > Yeah, that's true, it was more of a convenience to be able to test it easier in the integration test, otherwise I had to get into more hoops to know if the other node had sent some flags because a 0 value could mean no flags or flags not supported.

Ah I see.

If the non-test code itself wants to distinguish between no flags and no flags supported it can use `raft_append_entries_result->version` (although I suspect it won't be needed in practice).

For the test code in this specific PR, what do you think to just keep the:

```C
munit_assert_ullong(raft->leader_state.progress[1].feature_flags, ==,
                        RAFT_DEFAULT_FEATURE_FLAGS);
```

assertion below in the `replication/receiveFlags`? It's true that for now it won't distinguish the two cases, but it is still exercising the intended code paths and it distinguish the two cases once we add the first feature flag, so it's a short-lived glitch. On the other had it seems a bit odd to have to support the `RAFT_FLAGS_FLAGS` flag indefinitely for this one test that is only temporarily sub-optimal.

include/raft.h
freeekanayaka: After reading the reply to the `RAFT_FLAGS_FLAGS` comment, I assume the use case for this field is to tweak it in tests to exercise combinations of legacy and newer nodes?

src/flags.h
freeekanayaka: Anyway, this is minor, and I can see the rational of both approaches. I'm fine both ways.

include/raft.h
MathieuBordere: I can imagine it not being a compile-time constant, maybe some option is set at initialization and it changes the flags.

src/flags.h
MathieuBordere: I'll dump the `RAFT_FLAGS_FLAGS` it's just weird, not really useful and transient anyway.

freeekanayaka: One middle-ground approach could be to leave `RAFT_FLAGS_FLAGS` as it is just for this PR, but override/repurpose its bit by renaming `RAFT_FLAGS_FLAGS` to `RAFT_FLAGS_CHUNKED_INSTALL_SNAPSHOT` (or whatever name) in the follow-up PRs. That way we would use `RAFT_FLAGS_FLAGS` temporarily just for these preparation branches, and replace it with an actual feature once ready.

Unless I'm missing something I don't see us taking advantage of `RAFT_FLAGS_FLAGS` in any way in the future, so we might well get rid of it.

include/raft.h
freeekanayaka: > I can imagine it not being a compile-time constant, maybe some option is set at initialization and it changes the flags.

If that's the only reason we have for now, what do you think of making it a compile-time constant for now and turn it into a field only once we'll actually have a use case for dynamically configured features?

Note that in general for supporting dynamically configured features we might need specific logic (if/else), while if we stick to a compile-time constant `struct raft` can assume that all features are on locally, and only have logic to handle flags that are off remotely.

For the use case of tweaking it in `./test/integration`, I'm not entirely sure how much is useful, because to have it truly simulate a legacy `struct raft` we should have it run the same code paths/logic, which are typically not available in the current version of the code.

Perhaps to test that behavior we can rather feed `struct raft` with a manually crafted `struct raft_message`. For example a `struct raft_append_entries_result` with flags set to 0, and then check that the node sends out a single-chunked `struct raft_install_snapshot` instead of a multi-chunked one.

MathieuBordere:
> I'm wondering if we should include the `flags` field in all messages, not just `AppendEntriesResult`, for future proofing. For example, a follower will receive an `AppendEntries` message that will contain the leader's feature flags and perhaps modify its behavior accordingly. Same for `RequestVote`/`RequestVoteResult`.
> 
> I'm just thinking loud here, not sure.

I can see the use for it, it would complicate this PR and not have direct use though, a bit reluctant to add it now.

freeekanayaka:
> > I'm wondering if we should include the `flags` field in all messages, not just `AppendEntriesResult`, for future proofing. For example, a follower will receive an `AppendEntries` message that will contain the leader's feature flags and perhaps modify its behavior accordingly. Same for `RequestVote`/`RequestVoteResult`.
> > I'm just thinking loud here, not sure.
> 
> I can see the use for it, it would complicate this PR and not have direct use though, a bit reluctant to add it now.

Sure, I was meaning as separate work/PR. Wanted to gather feedback first though.

include/raft.h
freeekanayaka: Thanks. At this point what do you think of also converting this typedef from `uint64_t` to `unsigned long long`, for consistency with other types and to separate encoding and model concerns?

I voluteer for contextually improving `uv_encoding.c` to add boundary/validity checks, something we need to anyways since they are not there at the moment.

include/raft.h
```diff
@@ -264,8 +269,9 @@ struct raft_append_entries_result
     raft_term term;            /* Receiver's current_term. */
     raft_index rejected;       /* If non-zero, the index that was rejected. */
     raft_index last_log_index; /* Receiver's last log entry index, as hint. */
+    raft_flags flags;          /* Feature flags. */
```
freeekanayaka, line 272: Perhaps it would help a bit if we rename this field from `flags` to `features`? The word "flags" makes me think to specific options that apply the message. Renaming it "features" would make it a bit more explicit that this field contains the features the sending node knows about.

We might actually add a real message-specific "flags" in the future that is separate from the "features". Also, the `RequestVote` encoding already uses a "flags" field (see [here](https://github.com/canonical/raft/blob/529c5c83f44f32432131c43eee5b7bfd27a1df4d/src/uv_encoding.c#L90)) to encode message-specific flags. If we were to add the "features" field to also other messages in the future, this rename would avoid a clash, at least in terms of internal code/encoding.

freeekanayaka (approved):
Looks good to me.

I left a couple of other comments, but I'm ok either way.

include/raft.h
MathieuBordere: I kind of like the fixed width types, would prefer to keep as-is, if we ever need to store it in one of the reserved fields it also already has the desired size.

PR #439, opened 2023-06-22
configure: Use only pkg-config to detect lz4

freeekanayaka:
The lz4 upstream project has had pkg-config support since about 9 years:

https://github.com/lz4/lz4/commit/dc107107f73942574a0b2d1d9c1edd05b0ff1796

which is included in the r120 release from July 2014.

(the first lz4 release which isn't labeled with the rNNN pattern is v1.7.3 from November 2016, see https://github.com/lz4/lz4/tags?after=v1.8.0).

The oldest lz4 release still in the Ubuntu archive is r131, from 2015, see:

http://archive.ubuntu.com/ubuntu/pool/main/l/lz4/?C=M;O=A

which was included in Ubuntu 16.04 and sports pkg-config support.

It's safe to assume than any real-world consumer will have an lz4 version with pkg-config support.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -26,38 +26,7 @@ AC_ARG_ENABLE(lz4, AS_HELP_STRING([--disable-lz4], [do not use lz4 compression])
 # Allow not linking to liblz4 even if it's present.
 AC_ARG_WITH([lz4], AS_HELP_STRING([--without-lz4], [never link to liblz4]))
 
-# Thanks to the OpenVPN configure.ac file for this part.
-# If this fails, we will do another test next.
-# We also add set LZ4_LIBS otherwise linker will not know about the lz4 library
-PKG_CHECK_MODULES(LZ4, [liblz4 >= 1.7.1], [have_lz4="yes"], [LZ4_LIBS="-llz4"])
-if test "${have_lz4}" != "yes" ; then
-    AC_CHECK_HEADERS([lz4.h],
-                     [have_lz4h="yes"],
-                     [])
-    if test "${have_lz4h}" = "yes" ; then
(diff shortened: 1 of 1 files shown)


configure.ac
```diff
@@ -26,38 +26,7 @@ AC_ARG_ENABLE(lz4, AS_HELP_STRING([--disable-lz4], [do not use lz4 compression])
...
-             have_lz4="yes"
-         ],
-         [
-             AC_MSG_RESULT([system LZ4 library is too old])
-             have_lz4="no"
-         ]
-        )
-    fi
-fi
+PKG_CHECK_MODULES(LZ4, [liblz4 >= 1.7.1], [have_lz4="yes"], [have_lz4="no"])
```
MathieuBordere, line 29: `LZ4_LIBS` is not set anymore but it's used in `Makefile.am`. Or is there some magic that does it?

MathieuBordere:
I have probably added that part for a reason, can't remember exactly why of course ... :-)

configure.ac
freeekanayaka: `LZ4_LIBS` is set by the `PKG_CHECK_MODULES` macro, using what is reported by `pkg-config`.

The explicit set that was there in the code was there as a fallback, indeed it was set in the false branch. After the false branch was taken and `LZ4_LIBS` was set, the manual `if test "${have_lz4}" != "yes" ; then` fallback check would be executed, possibly leading to `have_lz4` being set to `yes`, and at that point `LZ4_LIBS` would remain set at the value set in the false branch of the `PKG_CHECK_MODULES` macro, and then be used in `Makefile.am`.

I'd say that the fallback `LZ4_LIBS` setting should have been performed inside the `if test "${have_lz4}" != "yes" ; then` fallback check, but the net effect is the same.

PR #445, opened 2023-06-27
Remove blocking barrier

MathieuBordere:
This is a draft PR that implements the approach to remove the (non)blocking barrier distinction discussed in #444 

I believe @cole-miller is not a huge fan of the approach. Alternatives would be:

- new segment format that encodes the index explicitly (more risky, think about downgrading raft versions, also need to be able to still handle old format segments)
- run a barrier on the first snapshot only, this carries a performance penalty as new writes will be blocked during the snapshot.
- ...

### DIFF
--- a/src/uv.c
+++ b/src/uv.c
@@ -364,6 +364,13 @@ static int uvLoadSnapshotAndEntries(struct uv *uv,
         goto err;
     }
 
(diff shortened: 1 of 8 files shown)


MathieuBordere:
Let me first fix the tests ... :-)

PR #446, opened 2023-06-27
Remove blocking barrier distinctions

MathieuBordere:
This is a draft PR that implements the approach to remove the (non)blocking barrier distinction discussed in https://github.com/canonical/raft/pull/444

I believe @cole-miller is not a huge fan of the approach. Alternatives would be:

- new segment format that encodes the index explicitly (more risky, think about downgrading raft versions, also need to be able to still handle old format segments)
- run a barrier on the first snapshot only, this carries a performance penalty as new writes will be blocked during the snapshot.
- ...


MathieuBordere:
Jepsen doesn't seem extremely happy https://github.com/canonical/jepsen.dqlite/actions/runs/5390663924/jobs/9786357309, taking a look before removing draft.

MathieuBordere:
Investigating the failures, it looks like raft has trouble decoding a configuration found in the raft log when taking a snapshot. All of the occurrences follow a truncation of the log though. I'm suspecting that removing the barrier before taking the snapshot, together with the truncation - that in turn also requires a barrier - has some unanticipated interaction.

src/uv_snapshot.c
```diff
@@ -578,6 +578,11 @@ static void uvSnapshotPutAfterWorkCb(uv_work_t *work, int status)
     struct uv *uv = put->uv;
     assert(status == 0);
     uv->snapshot_put_work.data = NULL;
+    /* The next entry in the first open segment will come after the snapshot
+     * entries. */
+    if (put->trailing == 0 && uv->prepare_next_counter == 0) {
+        uv->prepare_next_counter = 1;
```
freeekanayaka, line 584: It might be unrelated to the Jenkins failures, but after a very quick look at the PR's diff I noticed this change, and I'm not entirely sure it's enough. Could it be that there is a ready but yet unused `open-0` open segment that will be the one used after the snapshot is installed?

Just throwing it there in case it help, didn't really look much.

freeekanayaka:
> Investigating the failures, it looks like raft has trouble decoding a configuration found in the raft log when taking a snapshot. All of the occurrences follow a truncation of the log though. I'm suspecting that removing the barrier before taking the snapshot, together with the truncation - that in turn also requires a barrier - has some unanticipated interaction.

Perhaps it'd be a good idea to attach to this PR the Jepsen's artifacts containing the logs of one the failing tests, for future reference (since they'll get deleted at some point).

MathieuBordere:
My gut feeling is that due to the issue in #435 (where a second barrier callback doesn't fire when all active segments are involved in a barrier.), the log truncation didn't happen when a snapshot barrier was active covering up an issue that is being exposed now. But still investigating.

freeekanayaka:
> My gut feeling is that due to the issue in #435 (where a second barrier callback doesn't fire when all active segments are involved in a barrier), the log truncation didn't happen when a snapshot barrier was active covering up an issue that is being exposed now. But still investigating.

Could well be, but if that's the case, I'm wondering why such a bug was not surfaced by Jepsen also before the non-blocking barrier code was put in place. Or maybe our Jepsen suite is doing more nasty tests now that it was not doing before. Or we simply didn't examine some Jepsen failures before?

Anyway, logs should indeed shed some light on this.

MathieuBordere:
> > My gut feeling is that due to the issue in #435 (where a second barrier callback doesn't fire when all active segments are involved in a barrier), the log truncation didn't happen when a snapshot barrier was active covering up an issue that is being exposed now. But still investigating.
> 
> Could well be, but if that's the case, I'm wondering why such a bug was not surfaced by Jepsen also before the non-blocking barrier code was put in place. Or maybe our Jepsen suite is doing more nasty tests now that it was not doing before. Or we simply didn't examine some Jepsen failures before?
> 
> Anyway, logs should indeed shed some light on this.

yeah, I'm a bit reasoning on instinct here, I'm trying to gather proof now :-)

MathieuBordere:
What seems to happen is the following:

Jepsen has a snapshot trailing of `1024` and snapshot threshold of `128`.

Node takes snapshot at 2454
```
LIBRAFT   1688115651332445058 src/replication.c:1582 take snapshot at 2454
```
Data folder contains segment `0000000000000001-0000000000001408` and `open-1` containing 1K+ entries.

After the snapshot completes, the log is truncated, but only the closed segment is removed, the open segment, that was not closed, because we don't fire a barrier anymore before the snapshot, is left as-is.

Next time we restart, raft sees the open segment, no closed segments and the snapshot and erroneously puts all entries in the open segment after the snapshot leading to data corruption.

Will try to investigate what we can do, an explicit start index in an open segment sure sounds nice ... it just too tricky without it. Leaving the barrier before taking a snapshot and living with the extra complexity is an easier, and already tested option too...

MathieuBordere:
> edit: Or we could say, "Always leave at least the last closed segment", but then what's the point of the trailing parameter, it gets a bit fuzzy then.

I could live with this by the way, i'd rather have this than the blocking / non-blocking barrier stuff.

MathieuBordere:
Also, like @cole-miller mentioned, the introduction of `open-0` will also break a raft downgrade, so we could just as well move to a new segment format ...

PR #447, opened 2023-06-28
Invoke the new downstream checks workflow for PRs

cole-miller:
Let's see if this works!

### DIFF
--- a/.github/workflows/downstream.yml
+++ b/.github/workflows/downstream.yml
@@ -1,0 +1,95 @@
+name: Downstream checks
+
+on:
+  issue_comment:
+    types: [created, edited]
+
+if: ${{ github.event.issue.pull_request && contains(github.event.comment.body, 'please test downstream') }}
+
+jobs:
+  dqlite:
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Install apt deps
+        run: |
+          sudo apt-get update -qq
+          sudo apt-get install -qq automake libtool gcc make liblz4-dev libuv1-dev libsqlite3-dev
+
+      - name: Check out libbacktrace
+        uses: actions/checkout@v3
+        with:
+          repository: ianlancetaylor/libbacktrace
+          path: libbacktrace
+
+      - name: Install libbacktrace
+        run: |
+          cd libbacktrace
+          autoreconf -i
+          ./configure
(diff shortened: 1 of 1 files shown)


freeekanayaka:
> The latest revision implements what we discussed in [canonical/jepsen.dqlite#116](https://github.com/canonical/jepsen.dqlite/pull/116) -- you can trigger a run of the downstream tests by leaving a comment that contains the substring `please test downstream`, and similarly `please test jepsen`. (We can bikeshed the exact magic phrases.)

Sounds like a great start.

> Two problems remaining that I can think of:
> 
> * The jobs that are triggered in this way don't show up with the other checks at the bottom of the comment thread for the PR -- you have to go to the "Actions" tab to see them.

They don't even add an automatic comment after they finish right?

> * If you run downstream/Jepsen tests once, then fix something in your PR branch and push, those tests won't run again automatically. You have to leave another comment with the same magic string.
> 
> Are those showstoppers?

IMO, no. It's an improvement that we can start using. If it turns out we need something more sophisticated, things can be changed.

.github/workflows/downstream.yml
```diff
@@ -0,0 +1,78 @@
...
+        uses: actions/checkout@v3
+        with:
+          ref: refs/pull/${{ github.event.issue.number }}/head
+          path: raft
+
+      - name: Install raft
+        run: |
+          cd raft
+          autoreconf -i
+          ./configure --enable-debug --enable-uv --enable-sanitize
```
freeekanayaka, line 28: Perhaps `--enable-backtrace` too?

.github/workflows/downstream.yml
```diff
@@ -0,0 +1,78 @@
...
+          path: go-dqlite
+
+      - name: Test go-dqlite
+        env:
+          GO_DQLITE_MULTITHREAD: '1'
+        run: |
+          cd go-dqlite
+          go get -tags libsqlite3 -t ./...
+          go test -asan -v -race ./...
+          VERBOSE=1 DISK=${{ matrix.disk }} ./test/dqlite-demo.sh
```
freeekanayaka, line 63: I presume the `{{ matrix.disk }}` token expands to empty string and can be dropped?

freeekanayaka (approved):
Just a couple of minor comments.

PR #459, opened 2023-07-19
Turn off auto-recovery-by-default during integration tests

freeekanayaka:
This commit only changes the test suite.

It turns off the auto-recovery behavior of automatically re-trying to load the data directory in case `RAFT_CORRUPT` is hit by `raft_io->load()`.

Instead, tests now have to explicitly turn on auto-recovery if they want to enable that behavior.

This reduces the risk of inadvertently masking bugs, for example by having tests that only pass because auto-recovery is on, but would otherwise fail.

Note that auto-recovery is a best-effort and last-resort measure to prevent startup failures possibly caused by raft bugs that have left a broken state at shutdown. However it's something that we don't expect to happen, and this features comes at the cost of possible data loss. We should then normally turn it off in our tests, in order to catch bugs, and only turn it on in some tests in order to exercise the auto-recovery feature itself.

### DIFF
--- a/test/integration/test_uv_load.c
+++ b/test/integration/test_uv_load.c
@@ -285,7 +285,6 @@ struct snapshot
     do {                                                          \
         LOAD_VARS;                                                \
         SETUP_UV;                                                 \
-        raft_uv_set_auto_recovery(&f->io, false);                 \
         _rv = f->io.load(&f->io, &_term, &_voted_for, &_snapshot, \
                          &_start_index, &_entries, &_n);          \
         munit_assert_int(_rv, ==, RV);                            \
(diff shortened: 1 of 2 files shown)


freeekanayaka:
In the actual non-test code there is only a single line where the `uv->auto_recovery` is used: [here](https://github.com/canonical/raft/blob/87172dd56fc00ff120993b41eac99a406ddd50d9/src/uv.c#L456).

However, we have quite a few tests to cover that single switch, perhaps it'd be better to reduce them to one or two tests?

Code-coverage would remain unchanged, and we'd use those couple tests to prove that the feature works, rather than to exercise it for all cases where `RAFT_CORRUPT` is returned, which feels a bit of a combinatorial explosion.

freeekanayaka:
As a follow-up, I'd like to also make a similar change in Jepsen, where it's relatively more likely that bugs might get masked by auto-recovery being on.

freeekanayaka:
> > Also, does it make sense to do this for the dqlite tests as well?
> 
> I'd say yes.

Although, those are currently very unlikely to mask any bugs. So the most important tests would be raft itself and Jepsen.

freeekanayaka:
> > > In the actual non-test code there is only a single line where the uv->auto_recovery is used: [here](https://github.com/canonical/raft/blob/87172dd56fc00ff120993b41eac99a406ddd50d9/src/uv.c#L456).
> > > However, we have quite a few tests to cover that single switch, perhaps it'd be better to reduce them to one or two tests?
> > 
> > 
> > I haven't spent time in this part of the code, but surely the point of those tests is to pin down the behavior of auto-recovery under various circumstances, rather than line coverage as such?
> 
> Yes, but I think a middle ground should found. Let's say that we add a new test for a new condition that triggers a `RAFT_CORRUPT` error: should also a test be added that exercises the exact same scenario but with auto-recovery turned on? It feels a bit odd to me.
> 
> Since it's a binary switch, I think a handful of tests would be practically enough. Anyway, that'd be a separate PR.

I'd also add that the goal of these unit/integration tests is more about code coverage than "smoke" testing: ideally if we change some code, there should be at least a test failing, proving that the underlying code is needed and exemplifying why. But once there's a test or two failing, and the underlying code is fully covered, there's not much need to add more.

We have a lot of other cases with code paths that have a top-level `if` with a lot of lower-level code that can make the top-level `if` either take the true or the false branch: if we had to have tests for all possible logical combinations of the top-level if and the specific lower-level situations that yield the false or the true branch, then we'd really have lots of tests. Having a few tests that cover both the true and the false branch is **generally** enough (regardless of the underlying conditions that triggered them).  Hope the point is clear :)

freeekanayaka:
> I do think it's valuable to have _integration_ tests that are not focused on line coverage, because they're agnostic about points like how auto-recovery is implemented. But I don't dispute the point about the intent behind the existing tests.

These integration tests are decoupled from implementation details, in the sense that they only use the public API. However they are not completely oblivious of the implementation: when we decide to write a new integration test we do so because it covers a specific situation that the code deals with. The point I'm trying to make is that many existing tests cover effectively the **same** specific situation (triggering auto-recovery upon `RAFT_CORRUPT`) and so they are redundant.

PR #467, opened 2023-08-15
Don't convert to candidate while entries are being persisted, take 2

### LINKED ISSUE #386: recvAppendEntries: Assertion `r->state == RAFT_FOLLOWER || r->state == RAFT_CANDIDATE` failed

cole-miller:
Fixes #386, reviving the approach from #464 as the first part of the approach worked out in the discussion of #465.

I'm not entirely happy about adding the append_in_flight_count field to `struct raft` -- it would be more principled to query our raft_io for this information. On the other hand, adding a method for that is a more invasive change that breaks compatibility with older raft_io implementations. So I'm undecided on the best design here.

### DIFF
--- a/include/raft.h
+++ b/include/raft.h
@@ -749,7 +749,8 @@ struct raft
                 raft_id id;
                 char *address;
             } current_leader;
-            uint64_t reserved[8]; /* Future use */
+            uint64_t append_in_flight_count;
+            uint64_t reserved[7]; /* Future use */
         } follower_state;
         struct
         {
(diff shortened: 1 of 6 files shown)


MathieuBordere:
Looks good. If I'm being nitty I'd say that a separate commit for to initialize append_in_flight_count is a bit over the top.

MathieuBordere (approved):

MathieuBordere:
Just letting up a balloon. Does the same reasoning hold for a leader, i.e. should we not step down from Leadership when an append request is in flight?

PR #474, opened 2023-09-04
uv_fs: Don't do synced writes when emulating fallocate

MathieuBordere:
With the default segment size, and a blocksize of 4KB, the fallocate emulation code could perform 2048 synced writes leading to a lot of latency. Instead, sync the file once after writing without O_DSYNC. and reopen the file with O_DSYNC before giving the fd back to raft. Detect fallocate capabilities at startup.


MathieuBordere:
Fixes #476

PR #477, opened 2023-09-12
client: Only send timeout now message if persisted state is up-to-date.

MathieuBordere:
Receivers of the "timeout now" message ignore the message if their in-memory logs are not up-to-date or if they are persisting log entries. When the message is ignored, the sender will not retry it which can lead to failed leadership transfers that could have been avoided. Instead, let the leader only send the message when it knows the follower has persisted all log entries, based on the follower's match_index.

Fixes https://github.com/canonical/go-dqlite/issues/271. I can no longer reproduce that bug while it was easy to reproduce without the fix.

### DIFF
--- a/src/client.c
+++ b/src/client.c
@@ -435,7 +435,7 @@ int raft_transfer(struct raft *r,
 
     membershipLeadershipTransferInit(r, req, id, cb);
 
-    if (progressIsUpToDate(r, i)) {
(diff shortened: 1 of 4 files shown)


MathieuBordere:
please test downstream

MathieuBordere:
Tests look fine, I'll merge for now.

Issue #470, opened 2023-08-23
src/log.c:87: refsTryInsert: Assertion `next_slot->term != term' failed.


MathieuBordere:
Detected in Jepsen: https://github.com/canonical/jepsen.dqlite/actions/runs/5953700449/job/16148585347
[jepsen-data-bank-partition,disk-failure(2).zip](https://github.com/canonical/raft/files/12421760/jepsen-data-bank-partition.disk-failure.2.zip)

freeekanayaka:
> Ahh, and importantly:
> 
> * n3 is the leader in term 2, and its disk write for the entry at 3969 fails, causing it to step down
> 
> So I think the cleanup after the disk write failure must be buggy, missing a refcount decrement maybe

My 2c: when `struct raft_io` reports a disk write failure (i.e. `raft_io->append()`'s callback fires with a non-zero `status`), `struct raft` should somehow schedule a `raft_io->append` retry later in the future, rather than rolling back its own in-memory log and rely on the leader to resend everything and retrigger the whole behavior. Rolling back is always hard and it's easy to miss details, while retrying should be relatively straighforward (or at least it limits any rollback logic to the internals of `struct raft_io`).

In case the write failure is `ENOSPC` the follower `struct raft` could also take note of that and stop accepting new AppendEntries messages until the situation resolves.

cole-miller:
(fixed by #483)

Issue #430, opened 2023-06-07
v1.x RFC: pull based approach


freeekanayaka:
The `v0.x` series of libraft has been around for a few years now, and during that time we gained experience around its strengths and weaknesses.

I believe there's now enough data that we could aim at introducing a `v1.x` series with better long-term design, extensibility and maintainability.

I have a few ideas around this topic which I've been ruminating around for quite some time now, and I'd like to get feedback about them. I'll start by listing what I consider the aspect of the current design that would need to be improved and I'll then add possible approaches. It's not anything radical, but rather a change in the control flow that should however make things more flexible, composable and understandable.

##### `struct raft_io` and `struct raft_fsm`

The initial goal of the `struct raft_io` interface was to make it easy to implement the network and storage parts of `struct raft`, while keeping those concerns hidden from the point of view of `struct raft` consumers, which would get a reasonably friendly/intuitive API.

The same can be said of `struct raft_fsm`, which was meant to offer an easy way for `struct raft` consumers to plug in their application FSM.

However, we observed that there are actually several flavors/styles of FSMs and I/O that we want to support:  synchronous vs aynchronous FSM snapshots, in-memory vs disk-based FSMs, chunked snapshots, `io_uring`-style I/O (see below). All that puts a strain on `struct raft_io` and `struct raft_fsm` that must be made generic enough to accommodate all these various mechanics, and that leads to increased complexity.

##### push model

The current callback-based style (which I'll call "push model", because we have user's code that gets called by raft, instead of raft code that gets called by the user) was mainly inspired by `libuv` and other similar asynchronous frameworks in other languages (Javascript and Python).

That might be fine for writing asynchronous network clients or mostly-stateless web services (with request/response dynamics and state managed by a db). However it seems to complicate the control flow a bit for complex stateful engines like `struct raft`.

This is something that was discussed in `libuv` itself, however it is harder for them to move away from that because of backward compatibility, see https://github.com/libuv/leps/pull/3 and https://github.com/libuv/libuv/issues/6.

Interestingly, `io_uring` is instead based on a pull model, which is more flexible and efficient for asynchronous systems, because it gives control to the user and not to the framework.

#### Possible ideas for v1.x

I think both the two concerns above can be addressed by moving to a pull model.

[...]

PR #483, opened 2023-09-26
replication: Try to reinstate entries that have been truncated away

cole-miller:
Experiment with fixing #470

### DIFF
--- a/src/fixture.c
+++ b/src/fixture.c
@@ -706,11 +706,15 @@ static raft_time ioMethodTime(struct raft_io *raft_io)
 
 static int ioMethodRandom(struct raft_io *raft_io, int min, int max)
 {
-    struct io *io;
-    (void)min;
-    (void)max;
-    io = raft_io->impl;
-    return (int)io->randomized_election_timeout;
+    struct io *io = raft_io->impl;
+    int t = (int)io->randomized_election_timeout;
+    if (t < min) {
+        return min;
+    } else if (t > max) {
+        return max;
+    } else {
+        return t;
+    }
 }
 
 /* Queue up a request which will be processed later, when io_stub_flush()
--- a/src/fixture.c
+++ b/src/fixture.c
@@ -171,11 +171,10 @@ struct io
     unsigned disk_latency;                /* Milliseconds to perform disk I/O */
     unsigned work_duration;               /* Milliseconds to run async work */
 
-    struct
-    {
-        int countdown; /* Trigger the fault when this counter gets to zero. */
-        int n;         /* Repeat the fault this many times. Default is -1. */
-    } fault;
+    int append_fault_countdown;
+    int vote_fault_countdown;
+    int term_fault_countdown;
+    int send_fault_countdown;
 
     /* If flag i is true, messages of type i will be silently dropped. */
     bool drop[N_MESSAGE_TYPES];
@@ -186,40 +185,13 @@ struct io
     unsigned n_append;
 };
 
-/* Advance the fault counters and return @true if an error should occur. */
-static bool ioFaultTick(struct io *io)
+static bool faultTick(int *countdown)
 {
-    /* If the countdown is negative, faults are disabled. */
-    if (io->fault.countdown < 0) {
-        return false;
+    bool trigger = *countdown == 0;
+    if (*countdown >= 0) {
+        *countdown -= 1;
     }
-
-    /* If the countdown didn't reach zero, it's still not come the time to
-     * trigger faults. */
-    if (io->fault.countdown > 0) {
-        io->fault.countdown--;
-        return false;
-    }
-
-    assert(io->fault.countdown == 0);
-
-    /* If n is negative we keep triggering the fault forever. */
-    if (io->fault.n < 0) {
-        return true;
-    }
-
-    /* If n is positive we need to trigger the fault at least this time. */
-    if (io->fault.n > 0) {
-        io->fault.n--;
-        return true;
-    }
-
-    assert(io->fault.n == 0);
-
-    /* We reached 'n', let's disable faults. */
-    io->fault.countdown--;
-
-    return false;
+    return trigger;
 }
 
 static int ioMethodInit(struct raft_io *raft_io,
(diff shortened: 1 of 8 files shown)


cole-miller:
please test downstream

cole-miller:
please test downstream

cole-miller:
please test downstream

cole-miller:
Test results look encouraging!

cole-miller:
please test downstream

cole-miller:
please test downstream

calvin2021y:
no update for so long, any  strategic adjustment for this project?

freeekanayaka:
> no update for so long, any strategic adjustment for this project?

I'm sure Mathieu or Cole will be able to speak about that, but FWIW there's now also https://github.com/cowsql/raft, in which I've been fixing a few bugs surfaced by Jepsen tests and that I plan to incrementally move to a v1 version, with a design more or less described by #430 (backward compatibilty will be maintained for the foreseeable future though).

cole-miller:
canonical/raft is still maintained, but the dqlite team has been focused on a broad redesign of dqlite and libraft that aims to address some scalability problems with the current architecture. We're planning to implement that new design during the coming six-month cycle, and we'll publicize our planned changes to dqlite and libraft once the remaining details have been worked out. On the libraft side, our highest priority is making the library work well as a backend for dqlite, although we will maintain the API boundary between the two and avoid introducing any dqlite-specific code. We're not making it a goal to keep libraft backwards-compatible.

Basically, we're happy to see any individuals or projects using libraft, and we plan to keep maintaining it as a separate library, but the direction for it is going to be set by our plans for dqlite, which are going to dictate some big changes that likely won't be convenient for non-dqlite downstreams.

calvin2021y:
[cowsql/raft](https://github.com/cowsql/raft)  look very nice, I will try it. thanks you for this great project and always patient 
communication opportunities. 

For me reliability is more important than compatibility,  since I can always stop the cluster and dump data to upgrade into new version.

High performance comes after reliability, before compatibility.  

Thanks to the canonical /raft team for the hard work.

freeekanayaka:
> [cowsql/raft](https://github.com/cowsql/raft) look very nice, I will try it. thanks you for this great project and always patient communication opportunities.
> 
> For me reliability is more important than compatibility, since I can always stop the cluster and dump data to upgrade into new version.

The on-disk data format will likely not change in v1 of cowsql/raft (or change in minor backward-compatible ways), so no need for users to dump the data or perform data migration. What will change substantially is the C API, which will become more generic and flexible, with better support for disk-based FSMs, asynchronous FSMs, chunked snapshots, custom snapshotting approaches, etc. A compatibility layer will be kept around, so applications using the v0 API will have a very long code upgrade window to migrate from v0 to v1.

> High performance comes after reliability, before compatibility.

Reliability will be a focus too of cowsql/raft, with a requirement to keep all existing upstream and downstream tests (Jepsen, cowsql/dqlite, incus/lxd) always green throughout the development of v1. With the new API in place, there should be more room to improve performance as well, for example by using `io_uring` in very effective and efficient ways. It will take quite some time to get there though.

> Thanks to the canonical /raft team for the hard work.

You're welcome!

cole-miller:
Fixes #470. To sum up: libraft's in-memory log implementation can only track a single log entry for each (index, term) pair. We also need to keep log entries around as long as they're referenced by an `AppendEntries` message that's waiting to be sent. This leads to a situation where a leader steps down suddenly because its disk write for an entry fails, but it's still holding on to that entry in its in-memory log. Another node that received the entry from the original leader can win the next election and send the original leader that same entry, and the original leader will encounter a collision when it tries to append the entry to its log for a second time.

The fix is to have the original leader notice in this situation that the entry it's just received is already tracked by the in-memory log, and just increment the refcount of that existing entry.

src/replication.c
```diff
@@ -1171,6 +1171,21 @@ int replicationAppend(struct raft *r,
      * that we issue below actually completes.  */
     for (j = 0; j < n; j++) {
         struct raft_entry *entry = &args->entries[i + j];
+
```
MathieuBordere, line 1174: Reasoning out loud here:

The leader has tried to replicate a log entry say with data `D1` at term `T1` and index `I1`, replicates to a majority of followers doesn't hear back from any of them, disk write fails, and now we hit this logic and the leader tries to replicate a new log entry with data `D2` and term `T1` and `I1`, disk write succeeds and leader crashes. I think we now have a situation where there are different log entries for the same term and index.

src/log.h
```diff
@@ -27,6 +27,8 @@ struct raft_entry_ref
     raft_term term;              /* Term of the entry being ref-counted. */
     raft_index index;            /* Index of the entry being ref-counted. */
     unsigned short count;        /* Number of references. */
+    struct raft_buffer buf;      /* Private buffer for the tracked entry */
```
MathieuBordere, line 30: The comment is descriptive, but it would probably be more useful to add the rationale why `buf` and `batch` are added.

freeekanayaka:
FWIW, if it could serve as inspiration I've worked around this issue with this approach https://github.com/cowsql/raft/pull/101. Probably not ideal, but good enough for getting rid of the bug and the assertion failure (I'm planning a broader fix later).

src/log.c
```diff
@@ -480,6 +487,42 @@ static int ensureCapacity(struct raft_log *l)
...
+
+    for (slot = bucket; slot != NULL; slot = slot->next) {
+        if (slot->term == term) {
+            slot->count++;
+            l->back++;
+            l->back %= l->size;
+            entry = &l->entries[l->back];
+            entry->term = term;
+            entry->type = type;
+            entry->buf = slot->buf;
```
MathieuBordere, line 517: are `buf` and `batch` of the previous entry properly cleaned up when doing this?

cole-miller: They are cleaned up when the old entry is truncated away from the logical log, in removeSuffix. At this point in the code those fields are (logically) uninitialized.

src/fixture.c
```diff
@@ -262,6 +274,11 @@ static void ioFlushAppend(struct io *s, struct append *append)
         goto done;
     }
 
+    if (diskFaultTick(s)) {
```
MathieuBordere, line 277: `ioFaultTick` above provides similar functionality. Can you look if it's not too much pain to convert the tests that rely on `ioFaultTick` to simulate a disk failure to `diskFaultTick`?

cole-miller: Looking into this now

cole-miller: Done!

cole-miller: Not really, it would be better to have diskFaultTick and then a separate one to simulate network failures. The unified ioFaultTick is really not great imo and I would be happy to get rid of it.

cole-miller: This is implemented

PR #486, opened 2023-11-27
Initial log format rework

just-now:
- Removed static qualifier from tracef() to fix potential multithreading bug;
 - Added log levels.

### DIFF
--- a/README.md
+++ b/README.md
@@ -177,6 +177,9 @@ to detect corruptions that occurred during storage. It is therefore recommended 
 lz4 compression by means of the ```--disable-lz4``` configure flag.
 
 Detailed tracing will be enabled when the environment variable `LIBRAFT_TRACE` is set upon startup.
+The value of it can be in `[0..5]` range and reperesents a tracing level, where
+`0` means "no traces" emitted, `5` enables minimum (FATAL records only), and `1`
+enables maximum verbosity (all: DEBUG, INFO, WARN, ERROR, FATAL records).
 
 Notable users
 -------------
--- a/src/tracing.c
+++ b/src/tracing.c
@@ -1,43 +1,109 @@
 #include <inttypes.h>
 #include <stdlib.h>
+#include <string.h>      /* strstr, strlen */
+#include <sys/syscall.h> /* syscall */
 #include <time.h>
+#include <unistd.h> /* syscall, getpid */
 
+#include "assert.h" /* assert */
 #include "tracing.h"
 
-static inline void noopTracerEmit(struct raft_tracer *t,
-                                  const char *file,
-                                  int line,
-                                  const char *message)
+static inline void noopTracerEmit(struct raft_tracer *t UNUSED,
+                                  const char *file UNUSED,
+                                  unsigned int line UNUSED,
+                                  const char *func UNUSED,
+                                  unsigned int level UNUSED,
+                                  const char *message UNUSED)
 {
-    (void)t;
-    (void)file;
-    (void)line;
-    (void)message;
 }
 struct raft_tracer NoopTracer = {.impl = NULL,
                                  .enabled = false,
                                  .emit = noopTracerEmit};
 
+static inline const char *tracerShortFileName(const char *fname)
+{
+    static const char top_src_dir[] = "raft/";
+    const char *p;
+
+    p = strstr(fname, top_src_dir);
+    return p != NULL ? p + strlen(top_src_dir) : fname;
+}
+
+static inline const char *tracerTraceLevelName(unsigned int level)
+{
+    static const char *levels[] = {
+        "NONE", "DEBUG", "INFO", "WARN", "ERROR", "FATAL",
+    };
+
+    return level < ARRAY_SIZE(levels) ? levels[level] : levels[0];
+}
+
+static pid_t tracerPidCached;
+
+/* NOTE: on i386 and other platforms there're no specifically imported gettid()
+   functions in unistd.h
+*/
+static inline pid_t gettidImpl(void)
+{
+    return (pid_t)syscall(SYS_gettid);
+}
+
+static inline void tracerEmit(const char *file,
+                              unsigned int line,
+                              const char *func,
+                              unsigned int level,
+                              const char *message)
+{
+    struct timespec ts = {0};
+    struct tm tm;
+    pid_t tid = gettidImpl();
+
+    clock_gettime(CLOCK_REALTIME, &ts);
+    gmtime_r(&ts.tv_sec, &tm);
+
+    /*
+      Example:
+      LIBRAFT[182942] 2023-11-27T14:46:24.912050507 001132 INFO
+      uvClientSend  src/uv_send.c:218 connection available...
+    */
+    fprintf(stderr,
+            "LIBRAFT[%6.6u] %04d-%02d-%02dT%02d:%02d:%02d.%09lu "
+            "%6.6u %-7s %-20s %s:%-3i %s\n",
+            tracerPidCached,
+
+            tm.tm_year + 1900, tm.tm_mon + 1, tm.tm_mday, tm.tm_hour, tm.tm_min,
+            tm.tm_sec, ts.tv_nsec,
+
+            (unsigned)tid, tracerTraceLevelName(level), func,
+            tracerShortFileName(file), line, message);
+}
+
 static inline void stderrTracerEmit(struct raft_tracer *t,
                                     const char *file,
-                                    int line,
+                                    unsigned int line,
+                                    const char *func,
+                                    unsigned int level,
                                     const char *message)
 {
-    (void)t;
-    struct timespec ts = {0};
-    /* ignore errors */
-    clock_gettime(CLOCK_REALTIME, &ts);
-    int64_t ns = ts.tv_sec * 1000000000 + ts.tv_nsec;
-    fprintf(stderr, "LIBRAFT   %" PRId64 " %s:%d %s\n", ns, file, line,
-            message);
+    assert(t->level < TRACE_NR);
+
+    if (level >= t->level)
+        tracerEmit(file, line, func, level, message);
 }
+
 struct raft_tracer StderrTracer = {.impl = NULL,
                                    .enabled = false,
                                    .emit = stderrTracerEmit};
 
 void raft_tracer_maybe_enable(struct raft_tracer *tracer, bool enabled)
 {
-    if (getenv(LIBRAFT_TRACE) != NULL) {
+    const char *trace_level = getenv(LIBRAFT_TRACE);
+
+    if (trace_level != NULL) {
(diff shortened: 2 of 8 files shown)


src/tracing.c
```diff
@@ -1,43 +1,93 @@
...
 {
-    if (getenv(LIBRAFT_TRACE) != NULL) {
+    const char *trace_level = getenv(LIBRAFT_TRACE);
+
+    if (trace_level != NULL) {
+        tracer__pid_cached = getpid();
         tracer->enabled = enabled;
+
+        tracer->level = (unsigned)atoi(trace_level);
+        tracer->level = tracer->level < TRACE_NR ? tracer->level : TRACE_NONE;
```
MathieuBordere, line 91: What about having `TRACE_ERROR` as default level?

src/tracing.c
```diff
@@ -1,43 +1,93 @@
...
+                                const char *message)
+{
+    struct timespec ts = {0};
+    struct tm tm;
+    /* Use stack pointer instead of explicit thread id */
+    size_t sp = (size_t)&sp;
+
+    clock_gettime(CLOCK_REALTIME, &ts);
+    gmtime_r(&ts.tv_sec, &tm);
+
```
MathieuBordere, line 54: some example log line in the comment could be a nice add

README.md
```diff
@@ -177,6 +177,8 @@ to detect corruptions that occurred during storage. It is therefore recommended
 lz4 compression by means of the ```--disable-lz4``` configure flag.
 
 Detailed tracing will be enabled when the environment variable `LIBRAFT_TRACE` is set upon startup.
+The value of it can be in `0..8` range and reperesents a tracing level, where
+`0` means "no traces" emitted, and `8` enables maximum verbosity.
```
cole-miller, line 181: I'm used to `0..8` denoting the range `0 <= x < 8`, could this spell it out unambiguously?

.github/workflows/build-and-test.yml
```diff
@@ -16,8 +16,8 @@ jobs:
           - gcc
           - clang
         tracing:
-          - LIBRAFT_TRACE=1
-          - NOLIBRAFT_TRACE=1
+          - LIBRAFT_TRACE=8
+          - NOLIBRAFT_TRACE=8
```
cole-miller, line 20: Do we actually consult this env var anywhere?

src/tracing.h
```diff
@@ -15,16 +15,31 @@ extern struct raft_tracer NoopTracer;
...
+    TRACE_ALWAYS,
+    TRACE_FATAL,
+    TRACE_ERROR,
+    TRACE_WARN,
+    TRACE_NOTICE,
+    TRACE_INFO,
+    TRACE_DEBUG,
+    TRACE_CALL,
+    TRACE_NR
+};
```
cole-miller, line 39: Do we need this many levels? I feel like I'm not going to know which one to choose when I'm writing code without some sense of what the distinctions are.

src/tracing.c
```diff
@@ -1,43 +1,93 @@
...
+static inline void tracer__emit(const char *file,
+                                unsigned int line,
+                                const char *func,
+                                unsigned int level,
+                                const char *message)
+{
+    struct timespec ts = {0};
+    struct tm tm;
+    /* Use stack pointer instead of explicit thread id */
+    size_t sp = (size_t)&sp;
```
cole-miller, line 50: I assume `sp` here is some compiler magic? What information does printing the stack pointer add?

src/tracing.c
cole-miller: That seems like it would line up poorly with our existing use of `tracef` -- there are lots of call sites where it's not an error. Also we probably want `LIBRAFT_TRACE=1` to keep working

cole-miller: That seems like it would line up poorly with our existing use of tracef -- there are lots of call sites where it's not an error. Also we probably want LIBRAFT_TRACE=1 to keep working

.github/workflows/build-and-test.yml
MathieuBordere: It's a bit of a hack to test tracing on/off in a matrix https://github.com/canonical/raft/blob/a60437d43d566f5dab8e8e80d48f8609dc941a30/.github/workflows/build-and-test.yml#L45 `NOLIBRAFT_TRACE` does not enable tracing :-)

just-now:
```
[14:37:09] dqlite@raft git:(trace-fix)$ gg NOLIBRAFT_TRACE
.github/workflows/build-and-test.yml:20:          - NOLIBRAFT_TRACE=8
```

Let me delete `NOLIBRAFT_TRACE` at all. Any objections?

README.md
just-now: Agree, `[0, 8]` seems to look better.

src/tracing.c
MathieuBordere: If I'm not mistaken `LIBRAFT_TRACE=1` translates to `TRACE_ALWAYS`, which will still be shown if the default trace->level is `TRACE_ERROR`.

MathieuBordere: > Also we probably want LIBRAFT_TRACE=1 to keep working

Agree, `LIBRAFT_TRACE=1` should probably mean to trace everything.

src/tracing.h
just-now:
```
	/**
	 * represents an invalid trace level
	 */
	M0_NONE   = 0,

	/**
	 * special level, intended for
	 * printf-style debugging, should be removed from code before committing
	 */
	M0_ALWAYS = 1 << 0,

	/** system is not able to continue performing it's basic function */
	M0_FATAL  = 1 << 1,
	/**
	 * resource unavailable, no connectivity,
	 * incorrect value, etc.
	 */
	M0_ERROR  = 1 << 2,
	/**
	 * condition which requires some special
	 * treatment, something which not happens normally.
	 */
	M0_WARN   = 1 << 3,
	/** normal but significant code path */
	M0_NOTICE = 1 << 4,
	/** information about current system's state */
	M0_INFO   = 1 << 5,
	/**
	 * lower-level information to debug and analyse
	 * incorrect behavior
	 */
	M0_DEBUG  = 1 << 6,

	/**
	 * special level, used only to trace function calls
	 */
	M0_CALL   = 1 << 7,
```
We can try to narrow down the list.

src/tracing.c
just-now: Implication: 
```
#define Tracef(TRACER, ...) Tracef0(TRACER, TRACE_ALWAYS, __VA_ARGS__)
```

src/tracing.h
just-now: let's remove `ALWAYS`. It's too early for this class of logs.

just-now: `-NOTICE` and `-CALL`. In total:

```
	/**
	 * represents an invalid trace level
	 */
	M0_NONE  ,

	/** system is not able to continue performing it's basic function */
	M0_FATAL ,
	/**
	 * resource unavailable, no connectivity,
	 * incorrect value, etc.
	 */
	M0_ERROR  ,
	/**
	 * condition which requires some special
	 * treatment, something which not happens normally.
	 */
	M0_WARN  ,
	/** information about current system's state */
	M0_INFO  ,
	/**
	 * lower-level information to debug and analyse
	 * incorrect behavior
	 */
	M0_DEBUG  ,
```

cole-miller: Thanks, this list is much easier for me to understand. I like that there's a distinct level for transient debugging stuff! Could you add those comments to your PR for posterity?

src/tracing.c
cole-miller: Yeah, I think I'd be happy with just the PID for now, how do you feel @MathieuBordere?

src/tracing.c
MathieuBordere: Another suggestion is `LIBRAFT_TRACE=1` means 'LIBRAFT_TRACE=DEBUG` this is also in line with the current use of `tracef` as a very verbose logging system and would lead to
`#define Tracef(TRACER, ...) Tracef0(TRACER, TRACE_DEBUG, __VA_ARGS__)`

src/tracing.c
just-now: done.

src/tracing.c
just-now:
```
    gmtime_r(&ts.tv_sec, &tm);                                                                                                                                  
                                                                                                                                                                
    /*                                                                                                                                                          
      Example:                                                                                                                                                  
      LIBRAFT[182942] 2023-11-27T14:46:24.912050507 001132 INFO                                                                                                 
      uvClientSend  src/uv_send.c:218 connection available...                                                                                                   
    */                                                                                                                                                          
    fprintf(stderr,                                                                                                                                             
```
done.

src/tracing.h
just-now: done

src/tracing.c
just-now: Detailed tracing will be enabled when the environment variable `LIBRAFT_TRACE` is set upon startup.                                                             
The value of it can be in `[0..5]` range and reperesents a tracing level, where                                                                                 
`0` means "no traces" emitted, `5` enables minimum (FATAL records only), and `1`                                                                                
enables maximum verbosity (all: DEBUG, INFO, WARN, ERROR, FATAL records).

MathieuBordere (approved):
Looks good to me! I think we should follow up in a different PR where we set the default tracing level to `ERROR` and go over the existing `tracef` calls and adapt the levels where needed.

src/tracing.c
```diff
@@ -1,43 +1,109 @@
...
+static inline const char *tracerTraceLevelName(unsigned int level)
+{
+    static const char *levels[] = {
+        "NONE", "DEBUG", "INFO", "WARN", "ERROR", "FATAL",
+    };
+
+    return level < ARRAY_SIZE(levels) ? levels[level] : levels[0];
+}
+
+static pid_t tracer__pid_cached;
```
cole-miller, line 41: Is there a reason to keep the double underscore here?

src/tracing.h
```diff
@@ -15,16 +15,37 @@ extern struct raft_tracer NoopTracer;
...
+    /** Information about current system's state */
+    TRACE_INFO,
+    /**
+     * Condition which requires a special handling, something which doesn't
+     * happen normally
+     */
+    TRACE_WARN,
+    /** Resource unavailable, no connectivity, invalid value, etc. */
+    TRACE_ERROR,
+    /** System is not able to continue performing it's basic function */
```
cole-miller, line 42: Nit: `its`

cole-miller (approved):
Thanks, looks good except for a couple of tiny things!

src/tracing.h
just-now: fixed.

PR #492, opened 2024-01-25
Add raft_voter_contact()

ralight:
This returns the number of voting nodes that are recently in contact with the leader, to allow determining if the cluster is currently in a degraded / at risk state.

I'd be interested on your thoughts on this. At the moment it feels like not being able to find out this information is an omission.

### DIFF
--- a/include/raft.h
+++ b/include/raft.h
@@ -783,7 +783,9 @@ struct raft
             raft_index round_index;         /* Target of the current round. */
             raft_time round_start;          /* Start of current round. */
             void *requests[2];              /* Outstanding client requests. */
-            uint64_t reserved[8];           /* Future use */
+            uint32_t voter_contacts;        /* Current number of voting nodes we are in contact with */
+            uint32_t reserved2;             /* Future use */
+            uint64_t reserved[7];           /* Future use */
         } leader_state;
     };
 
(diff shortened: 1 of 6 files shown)


cole-miller:
Not speaking for the rest of the team, but this seems reasonable to me. I would like a note in the docstring that the return value is not to be relied on for correctness (since it could be out of date), and a couple of tests.

include/raft.h
```diff
@@ -783,7 +783,9 @@ struct raft
             raft_index round_index;         /* Target of the current round. */
             raft_time round_start;          /* Start of current round. */
             void *requests[2];              /* Outstanding client requests. */
-            uint64_t reserved[8];           /* Future use */
+            uint32_t voter_contacts;        /* Current number of voting nodes we are in contact with */
```
cole-miller, line 786: Please initialize this explicitly to 0 in raft_init, and reset it to 1 when entering leader state. (Without the second part, raft_voter_contacts could return a stale value from a previous term.)

cole-miller (changes requested):

ralight:
@cole-miller Thanks for the feedback. I've added the initialisation to 0 on init, reset to 1 on changing to a leader, and added a test which starts off with 3 nodes, kills off a node, and revives the node, with checks at each step. If you'd like more tests please give me some hints on the direction to go in.

MathieuBordere (approved):
Thanks for this, seems useful.

cole-miller (approved):
Thanks!

Issue #475, opened 2023-09-05
[suggestion] add lock/unlock function pointers to raft_io (and use them)


mdorier:
I ran into a tricky bug with my custom raft_io implementation, which took me a week to figure out by reading a lot of the raft code and the uv-based implementation. I'm posting a suggestion here as I remember you talked about refactoring the code.

A major difference between my implementation and the uv-based one is that my implementation has a coroutine-based I/O progress loop. This means that any blocking I/O function (such as writing something to a file, or sending an RPC) could cause a context-switch to the progress loop, giving it a chance to run something else, be it the ticker callback or the receive callback. I don't think this can happen with the uv-based implementation. (to be clear, this is not a multithread implementation in the sense that nothing actually executes concurrently. Simply, because of non-preemptive context-switching between coroutines, some code execution could be inserted during I/O calls).

An example of why this matters is the following, which was happening occasionally in my tests:

A raft process wants to become a candidate and start an election (say because the leader is a bit slow and a timeout was reached). Its `r->state` is set to `RAFT_CANDIDATE` and the `r->candidate_state` part of the union is allocated and used. The process enters `electionStart` in election.c. The assert that `r->state == RAFT_CANDIDATE` passes. It then calls `r->io->set_term`. In my custom implementation, this gives an opportunity to context-switch to the progress loop, which happens to have received a message from the leader, so the receive callback is called. Down the code path of this callback `r->state` is reset to `RAFT_FOLLOWER` and the `r->follower_state` part of the union is reinitialized, overriding `r->candidate_state`. At the end of the receive callback, my raft_io implementation context-switches back to the coroutine that was blocked on `r->io->set_term` inside `selectionStart`. The `assert(r->candidate_state.votes != NULL);` because it has garbage data (as it happens, `votes` in `candidate_state` is at the same location as the `raft_id id` in the `follower_state` part of the union, so you get something like `0x3` or whatever raft_id happens to be there). More importantly the coroutine then reaches `r->candidate_state.votes[i] = true`, which segfaults.

The solution I found was to use the `r->data` field to attach some data to the raft instance, including a mutex (custom mutex implementation that works with coroutines, in my case). Any client call to `raft_*` functions in my code is then surrounded by a lock/unlock of this mutex. Same for the call to the ticker callback and receive callback inside my raft_io implementation.

[...]

Issue #230, opened 2021-09-23
support send snapshot by patch mode


calvin2021y:
To avoid send large snapshot, we can use tools like bsdiff to generate patch for  snapshot,  and each node need keep some  snapshot index to generate the patch.

on all RPC results message need add  the node snapshot index list.  (number array)

When leader send snapshot install rpc,  use the follower's snapshot index list find a recent local snapshot, then made a bsdiff patch.  the install snapshot RCP should include the selected local  snapshot index.   

The follower will apply the patch with they own local snapshot, then restore the pached snapshot.

And a option to set how many snapshot should keep. 
 
This should be a options function so people can disable it,  if enabled they should provide a patch/apply_patch callback into FSM. (and this callback should be exec from uv_worker)

MathieuBordere:
Thanks for the suggestion. raft is surely going to need some way to decrease the size of the `InstallSnapshot` RPC in the future, another option would be to send the snapshots in chunks over the wire.

regarding `bsdiff`, in the manpage:
```
BSDIFF(1)                                                                      BSD General Commands Manual                                                                      BSDIFF(1)

NAME
     bsdiff — generate a patch between two binary files

SYNOPSIS
     bsdiff ⟨oldfile⟩ ⟨newfile⟩ ⟨patchfile⟩

DESCRIPTION
     bsdiff compares ⟨oldfile⟩ to ⟨newfile⟩ and writes to ⟨patchfile⟩ a binary patch suitable for use by bspatch(1).  When ⟨oldfile⟩ and ⟨newfile⟩ are two versions of an executable pro‐
     gram, the patches produced are on average a factor of five smaller than those produced by any other binary patch tool known to the author.

     bsdiff uses memory equal to 17 times the size of ⟨oldfile⟩, and requires an absolute minimum working set size of 8 times the size of oldfile.

SEE ALSO
     bspatch(1)

AUTHORS
     Colin Percival <cperciva@freebsd.org>

FreeBSD                                                                                May 18, 2003                                                                               FreeBSD
```

The fact that bsdiff would use memory 17! times the size of `oldfile` will be problematic, but the idea to generate a diff could be interesting.

PR #497, opened 2024-02-13
Mothball the project

cole-miller:


freeekanayaka:
Hey Cole, not really related to this PR, but I've noticed that issues have now been disabled for this repository. I assume the goal would be to avoid having people file new issues.

However, I think continuing to be able to access the existing issues would be valuable, since they contain discussions and considerations that might be relevant when understanding current code. Some of those issues are also referenced in the open issues that I see have been transferred to canonical/dqlite, and they contain additional information that is probably going to be useful when tackling them.

Since the rate of new issues opened by third parties is low, would it be possible to re-enable the issues section so that the old content is back online?

cole-miller:
Unfortunately just re-opening issues on this repository won't recover the issues that were transferred to canonical/dqlite. GitHub doesn't offer a way to nondestructively copy issues to another repository, and I wanted those issues to be present in canonical/dqlite since they're by and large still relevant to dqlite, so I transferred them there.

What do you think of the idea of adding a link to the canonical/raft README that points to https://github.com/canonical/dqlite/issues?q=is%3Aissue+label%3Atransferred-from-raft? That will list all the issues that were extant in canonical/raft just before I did the transfer.

cole-miller:
I've reopened issues for now so that everyone can browse the closed issues on this repository. The issues that I closed yesterday without transferring to canonical/dqlite are:

- https://github.com/canonical/raft/issues/430
- https://github.com/canonical/raft/issues/475
- https://github.com/canonical/raft/issues/230

freeekanayaka:
> Unfortunately just re-opening issues on this repository won't recover the issues that were transferred to canonical/dqlite. GitHub doesn't offer a way to nondestructively copy issues to another repository, and I wanted those issues to be present in canonical/dqlite since they're by and large still relevant to dqlite, so I transferred them there.
> 
> What do you think of the idea of adding a link to the canonical/raft README that points to https://github.com/canonical/dqlite/issues?q=is%3Aissue+label%3Atransferred-from-raft? That will list all the issues that were extant in canonical/raft just before I did the transfer.

I guess I didn't articulate this well. The issues that have been transferred are fine, since they have the same content and can be found in one way or another.

The problem I see is for the issues that were kept in this repository and are now not available anymore. For example: if one looks at some existing code and wants to see background discussions that led to it, now that can't be done anymore.

I'd suggest to just re-activate the Issues tab of this repository, which as far as I can see only has the downside of letting people post new content. But that should be rather rare and not a big deal.

cole-miller:
> I guess I didn't articulate this well.

No, it's my bad, I wasn't thinking about closed issues at all. You're right that it makes more sense to keep the issues open for that reason.

MathieuBordere (approved):

